PR #13, opened 2015-09-24
Dev/jim/jsonfixes

jimtarber:
JSON fixes for the problems in Mantis #2757 and #2883, and a few other unreported issues (see individual commits) as well as enhanced the code to support multiple leading BOMs (reported in Mantis #2882).


zauberparacelsus:
Somebody set us up the BOM!


...sorry, couldn't resist XD

PR #15, opened 2015-09-24
Misc cleanup

cinderblocks:
Can't run it yet, so these are just various small fixes I poked at that should be safe.


mdickson:
Nice catches!  I know there is more cleanup that can be done based on the warning output from Monodevelop/Xamarin.  Good to see changes for this kind of stuff.

ddaeschler:
Awesome! Thanks for the fixes and cleanup

PR #17, opened 2015-09-24
Dead store and unused entity removal

cinderblocks:
None of these variables are ever read or removed classes ever used. There are still more, but this is a nice start.


ddaeschler:
Just reviewed this and it looks good to me! Thanks!

PR #19, opened 2015-09-28
Added support for Mono's implementation of SQLite.

kf6kjg:
Mono ships its own implementation of SQLite: `Mono.Data.Sqlite`
Note the casing: Sqlite instead of SQLite.  This is the same across all the classes and functions, hence a pile of `#if __MonoCS__` tests.

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/StateManager.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/StateManager.cs
@@ -39,9 +39,14 @@ using InWorldz.Phlox.Serialization;
 using System.IO;
 using log4net;
 using System.Reflection;
+#if __MonoCS__
+using Mono.Data.Sqlite;
+#else
 using System.Data.SQLite;
+#endif
 using OpenSim.Region.Framework.Scenes;
 using OpenSim.Framework;
+
 
 namespace InWorldz.Phlox.Engine
 {
@@ -60,7 +65,11 @@ namespace InWorldz.Phlox.Engine
         /// <summary>
         /// The connection to the databases is kept open for the duration of the application
         /// </summary>
+        #if __MonoCS__
+        private SqliteConnection _connection;
+        #else
         private SQLiteConnection _connection;
+        #endif
 
 
         private DateTime _lastReport = DateTime.Now;
@@ -140,8 +149,12 @@ namespace InWorldz.Phlox.Engine
 
         private void SetupAndOpenDatabase()
         {
-            _connection = new SQLiteConnection(
-                String.Format("Data Source={0}", this.GetDBFileName()));
+            #if __MonoCS__
+            _connection = new SqliteConnection(String.Format("Data Source={0}", this.GetDBFileName()));
+            #else
+            _connection = new SQLiteConnection(String.Format("Data Source={0}", this.GetDBFileName()));
+            #endif
+
 
             _connection.Open();
 
@@ -149,7 +162,11 @@ namespace InWorldz.Phlox.Engine
 										        "FROM sqlite_master " +
 										        "WHERE type = 'table' AND tbl_name='StateData';";
 
+            #if __MonoCS__
+            using (SqliteCommand cmd = new SqliteCommand(INITIALIZED_QUERY, _connection))
+            #else
             using (SQLiteCommand cmd = new SQLiteCommand(INITIALIZED_QUERY, _connection))
+            #endif
             {
                 long count = (long)cmd.ExecuteScalar();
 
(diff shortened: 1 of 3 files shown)


mdickson:
I don't recall needing to use a special version of SQLite in the past. I thought the DLL included both native and portable code. But I can revisit and check it.  I'd hate to see the need for all the case maping #if changes if we don't absolutely need it.

kf6kjg:
Agreed.  What prompted this was that while the code compiled fine using `System.Data.SQLite`, it didn't have a valid SQLite DLL and so crashed at runtime.  Since Mono includes its own, I figured it was easier to just use that than to try and find and require people to go get the relevant DLL.

mdickson:
Well we can check the DLL into the project if needed.  I know I fixed this once before I just dont remember how I did it.   What you will need to do is have the system .so installed on the linux side so the DLL can find and call into that. 

We can always back this out when we have a solution for the portable DLL.  I just wanted to say I'd worked this out once before I just cant remember what the solution was.  LOL. Need to jog some brain cells and reconstruct what I did.

kf6kjg:
LOL, that's Ok.  I do remember seeing some scattered `#if __MonoCS__` mostlye in `Util.cs` locking out some Kernel32 calls.

And, apparently, I missed making the changes to `Scene.cs` a separate pull request: I'd found some more kernel32 calls that needed blocking out.  No implementation, so I can only hope the watchdog at least sorta works, but with this and some upcoming patches I did get a region booted and connected to our local grid.  Can see it on the map - though the maptile isn't rendered because of missing the OpenJPEG DLL or .so. Unable to TP into it, most likely because I'm hacking the connection: I'm attempting to run my region with all the ports tunneled through SSH to the actual grid...  Yeah, it's not happy about that!  But it's running! :D

mdickson:
Cool!  How did you get past the physics initialization? 

Interesting... you should be able to get at least the grid services running in Mono in thinking this through.  I'll make some time tonight and give some of that a go.  That should help to flesh out the platform dependencies a bit. I have been building on both Mono/Linux and Xamarin/Mac so it would be nice to address both platforms.

kf6kjg:
For how I got past the physics implementation, see https://github.com/InWorldz/halcyon/pull/20

kf6kjg:
During testing of #20, I replicated the core situation  hat stimulated me into creating this patch and got the crash output:
```
APPLICATION EXCEPTION DETECTED: System.UnhandledExceptionEventArgs

Exception: System.DllNotFoundException: System.Data.SQLite.dll
  at (wrapper managed-to-native) System.Data.SQLite.UnsafeNativeMethods:sqlite3_config (int,System.Data.SQLite.SQLiteLogCallback,intptr)
  at System.Data.SQLite.SQLite3.StaticIsInitialized () [0x00025] in <filename unknown>:0 
  at System.Data.SQLite.SQLiteLog.Initialize () [0x00000] in <filename unknown>:0 
  at System.Data.SQLite.SQLiteConnection..ctor (System.String connectionString) [0x0000e] in <filename unknown>:0 
...
```

kf6kjg:
If not the patch I made or a DLL-based fix, then a simple wrapper class could be written that makes everything consistent and internally handles the Mono-vs-System styles.

ddaeschler:
@mdickson is working on restoring the native components that he believes are missing on Linux to fix the functionality there.

kf6kjg:
@mdickson How's progress?

mdickson:
I have the openjpeg DLL's generated but the default versions that are released for SQLite are not cross platform.  I may need to rebuild from source to enable platform detection.    I was sidetracked for a while without a windows machine to test on but I have that fixed now.

An alternative is to switch to the Mono SQLite provider for everything since it will also work on Windows.   That would remove the need for the special case code w/ifdefs if we are comfortable using that provider across the board.  Looking for feedback on that.  BTW thats the approach used in OpenSim.

ddaeschler:
I have a hesitation about switching to the mono implementation since it is completely untested on the InWorldz grid and some pretty critical code currently relies on the existing implementation. Script state saves are one example and we have thousands of state files saved for all different regions across the grid. It is difficult to know what kind of impact (performance, etc) switching to the mono implementation would have.

I'm open for more input, but I'd rather keep what we have rather than rock the boat. Maybe @kf6kjg could assist with the cross platform DLL builds?

mdickson:
http://www.codeproject.com/Articles/821149/Universal-System-Data-Sqlite-binary-for-NET-and-Mo

URL that describes the build process for the DLL's.

kf6kjg:
Sounds like and adventure.  From the comments it looks like OSX might be an
issue.

It also looks like while he's packaging the dll with the release, he's
relying on the Linux ssytem to have SQLite preinstalled.

I'd have to dig further, but this might mean that Halcyon may need a build
system that can check for and download precompiled binaries from a central
storage location a la the client build process...  What do you think?

kf6kjg:
Ok, so I got it working on Linux, see https://github.com/kf6kjg/halcyon/tree/fix_mono_sqlite_trial2

However I've been unable to test on Windows 7 building with VS2015: no matter whether it is with my edition of the DLL or with the master branch's edition I get an error about assembly version conflicts on System.Data.SQLite.dll.  Any of you have issues getting a build to run with VS2015 under Win7?

kf6kjg:
Closing in favor of #48.

PR #20, opened 2015-09-28
Return of basicphysics

kf6kjg:
Everyone welcome our long-lost friend, BasicPhysics!

While BasicPhysics is nearly useless, it's the only tool to use when no other physics engine is available or working. 
It at least allows you to test that all the other code is running right!

This is important mainly on Linux where we don't at this time yet have a working PhysX implmentation.

### DIFF
--- a/OpenSim/Region/ClientStack/RegionApplicationBase.cs
+++ b/OpenSim/Region/ClientStack/RegionApplicationBase.cs
@@ -111,7 +111,7 @@ namespace OpenSim.Region.ClientStack
             m_httpServer.HostName = m_networkServersInfo.HostName; 
             MainServer.AddHttpServer(m_httpServer);
 
-            m_log.Info("[REGION]: Starting HTTP server");
+            m_log.InfoFormat("[REGION]: Starting HTTP server on {0}:{1}", m_httpServer.HostName, m_httpServerPort);
             m_httpServer.Start();
 
             // If specified configure an ssl server for use as well
--- a/OpenSim/Region/Physics/BasicPhysicsPlugin/BasicPhysicsPlugin.cs
+++ b/OpenSim/Region/Physics/BasicPhysicsPlugin/BasicPhysicsPlugin.cs
@@ -31,189 +31,223 @@ using OpenMetaverse;
 using OpenSim.Framework;
 using OpenSim.Region.Physics.Manager;
 
-namespace OpenSim.Region.Physics.BasicPhysicsPlugin
-{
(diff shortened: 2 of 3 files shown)


kf6kjg:
I'm not 100% confident that everything was implemented correctly, but assuming #19 and this are both merged, you should be able to bring a Linux install online.  I need to clean my local repository of uncommitted files and rebuild from scratch, double checking my assumptions, but I think I got everything.

ddaeschler:
I don't want to merge #19 yet until we take a look at why the universal binary isn't working for SQLite as to avoid using two different libraries depending on the platform. However, this PR is unobtrusive and I'm going to give it a simple shakedown and if everything looks good I'll merge it.

ddaeschler:
One issue:

In the prebuild.xml included in the PR, the target directory for the output is wrong. Once I fixed this the region started.

<Configuration name="Debug" platform="x64">
+        <Options>
+          <OutputPath>../../../../bin/</OutputPath>
+        </Options>
+      </Configuration>
+      <Configuration name="Release" platform="x64">
+        <Options>
+          <OutputPath>../../../../bin/</OutputPath>
+        </Options>
+      </Configuration>

should be 

<Configuration name="Debug" platform="x64">
+        <Options>
+          <OutputPath>../../../../bin/Physics</OutputPath>
+        </Options>
+      </Configuration>
+      <Configuration name="Release" platform="x64">
+        <Options>
+          <OutputPath>../../../../bin/Physics</OutputPath>
+        </Options>
+      </Configuration>

ddaeschler:
When you get a chance, please patch that issue in your PR and I will apply it.

kf6kjg:
Ah, I knew there was a likelihood of there being a miss somewhere: had not
enough time this morning before breakfast to do a complete cleanout and
test.  OK, I will fix that as soon as I can.

ddaeschler:
No problem. Part of my job now is to test branches/PRs before merging to make sure stuff is working.

kf6kjg:
Done, with also a style correction.  It looks like the default style is Visual Studio's.  I went and looked up Prebuild to see if it supported adding the MonoDevelop-specific rules for code formatting so that the developer (me) didn't have to reset that every time `runprebuild.sh` is executed. It doesn't, near as I can tell. Prebuild also seems to be a dead project: no movement since 2010.  Seems they figured that everyone's settled on MSBuild files and prebuild wasn't needed any more.

ddaeschler:
Start up! Thanks!

PR #27, opened 2015-10-03
Zauber halcyon 3

zauberparacelsus:
This branch extends adds implementations for several new functions: 

list iwReverseList(list src, integer stride);
Returns a reversed version of the list src.  The stride argument allows it to preserve the ordering of elements within strides.

string iwReverseString(string src);
Returns a reversed version of the string src.

list iwSearchInventory(integer type, string pattern, integer matchType);
Returns a list of all inventory items of the specified inventory type, if they match the specified pattern.  Follows the semantics of iwMatchString, and if pattern is an empty string, it will simply match every item of the specified type.

list iwSearchLinkInventory(integer link, integer type, string pattern, integer matchType);
As iwSearchInventory, but it can target another link within the same linkset.

integer iwIntRand(integer max);
Returns a random positive integer value, up to and including the specified maximum.

integer iwIntRandRange(integer min, integer max);
Returns a random integer value, ranging from minimum to maximum.

float iwFrandRange(float min, float max);
Returns a random float value, ranging from minimum to maximum.

list botSearchBotOutfits(string pattern, integer matchType, integer start, integer end);
As botGetBotOutfits, but it will return a filtered list of bot outfits with names matching a specified pattern, according to the semantics of iwMatchString.  If pattern is blank, then every name is matched.  The start and end values allow the function to only return a subset of the results, useful if a script's free memory is very low, or if the list happens to be very large.

In addition, llGetEnv() and llGetObjectDetails() have been extended to recognize several parameters, though some are unimplemented.

For llGetEnv(), I have added support for the following arguments: simulator_hostname, agent_limit, region_product_name, region_product_sku, estate_id, estate_name, region_start_time, and region_cpu_ratio.  Note that region_product_name and region_product_sku will return different results than in Second Life, due to InWorldz/Halcyon having different region products.  The region_cpu_ratio argument will simply always return 1 for the time being.

For llGetObjectDetails(), I have extended it to recognize OBJECT_RENDER_WEIGHT, OBJECT_HOVER_HEIGHT, OBJECT_BODY_SHAPE_TYPE, OBJECT_LAST_OWNER_ID, and OBJECT_CLICK_ACTION.  Note the following caveats, however:
* OBJECT_RENDER_WEIGHT is not supported by Halcyon at this time, so it returns 0.
* OBJECT_HOVER_HEIGHT returns 0.  This parameter does not refer to the Hover property in an avatar's shape as one might assume, but rather to a completely different feature in Second Life.
* OBJECT_BODY_SHAPE_TYPE returns 0.0 for female shapes and 1.0 for male shapes, instead of 0 and 255 as you would get from iwGetAppearanceParams().  This is to match Second Life behavior.


zauberparacelsus:
I forgot to note that I also extended iwMatchString to recognize the IW_MATCH_EQUAL and IW_MATCH_INCLUDE constants.

ddaeschler:
Passed code review. Looks good.

zauberparacelsus:
woot \o/

PR #30, opened 2015-10-04
Zauber halcyon 4

zauberparacelsus:
This branch contains a fix for an issue where script sensors, Highlights/Beacons, and Top Scripts would only show scripted objects if there were scripts in the root prim, ignoring scripted objects if their scripts were only in the child prim(s) and not in the root prim.  Due to these changes, more scripted objects will be revealed to people using the tools above when they go searching for scripted objects.

Also included is a correction to llGetObjectDetails, where OBJECT_LAST_OWNER_ID and OBJECT_CLICK_ACTION were not handled when the function targeted an avatar.

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
@@ -12728,6 +12728,13 @@ namespace InWorldz.Phlox.Engine
                         if (value == 255) ret.Add(1f);
                         else ret.Add(0f);
                         break;
+                    case ScriptBaseClass.OBJECT_LAST_OWNER_ID:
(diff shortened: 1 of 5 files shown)


zauberparacelsus:
Hold up, I found a bug or two to take care of...

zauberparacelsus:
Fixed two bugs.  The first fixes a bug that caused the IsScripted property to not get set to false when you removed scripts from the child prim.  The other fixes a bug where adding scripts to a child prim would not trigger a prim update in the root if the addition of those scripts changed the root prim's IsScripted property.  The update is necesary in order for the object to show up with script beacons and highlights.

ddaeschler:
After a code review this needs some more work to be merged. We're discussed privately the best way forward on this.

PR #31, opened 2015-10-04
Fixes for objects with scripted child prims

zauberparacelsus:
Makes some fixes so that objects which contain scripts in the child prims (but not the root prim) will properly show up as scripted objects.  This allows objects with scripts in the child parts to show up in Top Scripts and script sensors, along with being able to show up in viewer when script beacons or highlights are activated.  Also re-added the fix to GetAgentDetails to recognize OBJECT_LAST_OWNER_ID and OBJECT_CLICK_ACTION.


ddaeschler:
Code review looks good!

PR #32, opened 2015-10-05
Issue 24 purge folder fixed

kf6kjg:
Should be ready to go, though I'm pretty sure that the Cassandra method `InventoryStorage::PurgeFolderInternal()` isn't live-tested - the only place it was previously called was from `InventoryStorage::PurgeFolders()` which only seems to be exercised by a unit test: `UnitTests::TestMultiPurgeFolder()`.  There is also a call through the apparently unused method `CheckedInventoryStorage::PurgeFolders()`.

### DIFF
--- a/OpenSim/Region/Framework/Scenes/Scene.PacketHandlers.cs
+++ b/OpenSim/Region/Framework/Scenes/Scene.PacketHandlers.cs
@@ -628,10 +628,10 @@ namespace OpenSim.Region.Framework.Scenes
                         folderID, parentID, remoteClient.Name);
             }
             
-        }      
+        }
         
         /// <summary>
-        /// This should delete all the items and folders in the given directory.
+        /// This should recursively delete all the items and folders in the given directory.
         /// </summary>
         /// <param name="remoteClient"></param>
         /// <param name="folderID"></param>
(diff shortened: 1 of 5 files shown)


OpenSim/Region/Framework/Scenes/Scene.PacketHandlers.cs
```diff
@@ -680,6 +680,56 @@ public void HandlePurgeInventoryDescendents(IClientAPI remoteClient, UUID folder
...
+                    return;
+                }
+
+                //make sure the folder is either the lost and found, the trash, or a descendant of it
+                if (!CheckFolderHeirarchyIsAppropriateForPurge(folder, userProfile))
+                {
+                    m_log.ErrorFormat(
+                        "[AGENT INVENTORY]: Not purging descendents of {0} for user {1}, folder is not part of a purgeable heirarchy",
+                        folderID, remoteClient.Name);
+                }
```
ddaeschler: Oops.. Fall through here with no early return.. We'd complain about the hierarchy but then never break out.

kf6kjg: Wow didn't see that!  Both HandlePurgeInventoryFolder and HandlePurgeInventoryDescendents have are missing that same early return!

ddaeschler: Haha. Well in fairness I probably made that particular mistake, so if you patch them both up I'll be very grateful.

kf6kjg:
Fixed!

ddaeschler:
Looks good!

PR #33, opened 2015-10-05
Fixed #25: Under MySQL emptying the trash does not purge folders, only items.

### LINKED ISSUE #25: Under MySQL emptying the trash does not purge folders, only items
This seems specific to the MySQL storage implementation as I was not able to replicate the bug on the live InWorldz grid.

Here's how to test:
1. Create a new folder in your inventory. It will be named "New Folder".
2. Delete the new folder.
3. Observe that the "delete" has moved the folder to the Trash, and the "New Folder" is sitting right at the top. This is correct and expected.
4. Right-click on the Trash and select "Empty Trash".
5. Observe that the everything in the Trash disappears from your inventory. This is correct and expected: the client removes the items and folders from its own view of inventory.
6. Relog.
7. Open Inventory and expand your Trash folder.
8. Observe to see if the "New Folder" is in the Trash.

If step 8 shows the "New Folder", then the bug is still in effect.

Replicated with the following setup:
```
Firestorm 4.7.3 (47323) Aug 17 2015 18:14:04 (Firestorm-Releasex64) with OpenSimulator support
```
On current Halcyon compilation under Mono on Linux, as well as the 0.9.11 R5513 binary release of InWorldz.

(Side note: this is a fork of issue #23, as that issue was found to be covering at least 3 different problems at the same time.)

kf6kjg:
After all the work on 24, this seems like it was too easy.  Just had to add the logic that it needed to go do a recursive cleanup.

### DIFF
--- a/InWorldz/InWorldz.Data.Inventory.Cassandra/LegacyMysqlInventoryStorage.cs
+++ b/InWorldz/InWorldz.Data.Inventory.Cassandra/LegacyMysqlInventoryStorage.cs
@@ -150,7 +150,7 @@ namespace InWorldz.Data.Inventory.Cassandra
 
         public void PurgeFolderContents(InventoryFolderBase folder)
         {
-            _impl.deleteItemsInFolder(folder.ID);
+            _impl.deleteFolderContents(folder.ID);
         }
 
         public void PurgeFolder(InventoryFolderBase folder)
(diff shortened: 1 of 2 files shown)


ddaeschler:
That sounds good to me. I'll merge this when you've made the change.

kf6kjg:
So done. :)

ddaeschler:
Merged manually due to conflicts. Thanks!

PR #35, opened 2015-10-06
Zauber halcyon 5

zauberparacelsus:
This has a fix for a persistence that was uncovered, which I had inadvertently caused with my changes for checking if an object was scripted.  Also included are implementations of the four following script functions:

list iwListRemoveElements(list src, list elements, integer count);
list iwListRemoveDuplicates(list src);
iwStartLinkAnimation(integer link, string anim);
iwStopLinkAnimation(integer link, string anim);


ddaeschler:
Code reviewed. Looks good to go

PR #36, opened 2015-10-06
Replaced the now obsolete SVN-centric version info batch file.

jimtarber:
Replaced the old batch file to produce the SVN version info with build
date-based single file that is auto-updated on any full rebuild.


jimtarber:
I've changed the auto-versioning completely after a discussion with Dave. The first pass was an attempt to get something together we could use in time for our first production rollout of all these changes (Oct 19). However, he pointed out that some copies may not preserve the file timestamp and we could get some unexpected versions.  I have replaced the filestamp-based code with a new version which reads a version number that is stored in the file assembly (for OpenSim.Framework.Servers which is common to all services), and a revision number that is auto-generated by the build process.  It is using the default which is the number of days since 2000. (The fourth field is seconds since midnight.)  The previous InWorldz release was R5720, and today's auto-revision happens to be 5758, so that could not have worked out better. I've updated it to use the day count and we'll have a new revision number each day.  (Individual builds will also automatically be visible if you inspect the version info in the .DLL file.)  For production releases, or any binary transmission, there is no need to manually do anything unless you change the version number itself (currently 0.9.18 in the Halcyon builds).  I could have put this in a separate AssemblyInfo.cs file but there are so many of those and they are not maintained very well. I thought it best to keep the one version we need to update in the same place it has always been, at the top of the file.

PR #37, opened 2015-10-06
Dev/mdickson/rendermaterials classes

mdickson:
Base classes used for RenderMaterials support.  Includes getter/setters for individual parameters.  Also test cases for serialization as OSD which is needed by the CAP and a binary serialization which may be useful for SQLifying the data.

### DIFF
--- a/OpenSim/Framework/RenderMaterials.cs
+++ b/OpenSim/Framework/RenderMaterials.cs
@@ -1,0 +1,380 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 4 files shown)


mdickson:
Going to rework this a bit to use protobuf serialization for the byte arrays and simplify the container class.

mdickson:
Updates complete. Should be set to review and merge.

ddaeschler:
Looks good. Going to merge but check the hard tabs in a future revision. Looks like your editor is waffling between tabs and spaces.

PR #43, opened 2015-10-09
Issue 34 mysql folder versioning

kf6kjg:
Verified that all code paths properly update the version.

Also checked and implemented relevant logic from the Cassandra implementation - but especially not the redundancy I found there.

Especially note the changes around the delete*Folder* methods: I consolidated the database connections so that the server wasn't having to create and destroy potentially thousands of database connections any time someone tried to empty the trash or purge a folder.

That last also enabled me to make that process transactional - it either happens in its entirety or not at all.


ddaeschler:
Code reviewed. Looks good!

PR #45, opened 2015-10-10
Change to implicit animation perms for bots

zauberparacelsus:
This change allows animation perms to be implicitly granted for bots when the object requesting animations is owned by the bot's owner.  This change is to allow bots to be animated by scripted objects that have not been coded to handle bots.

I would like to have extended this so that if a bot is seated on an object, animation perms are implicitly granted for any object owned by the same owner as whatever the bot is sitting on, however that would require me to change an area of the codebase that I have not yet familiarized myself with.


ddaeschler:
@jimtarber can you review this one? You're more familiar with the implicit/perms side

jimtarber:
Looks good to me. The additional support for requests from scripts in other objects will be a lot easier to do once I finish avatar-as-a-prim support, so we can follow up with that then.  In the meantime, this is ready to merge.

PR #47, opened 2015-10-11
Ll get agent info flying sitting fix

zauberparacelsus:
Applies a workaround to an issue that caused llGetAgentInfo to indicate that an avatar was flying when they were sitting, triggered by the avatar trying to sit while they are flying, and then llGetAgentInfo being called upon the CHANGED_ANIMATION trigger firing in the changed event.


jimtarber:
This change looks good. It's an improvement to some pretty ugly code in there. But we can improve that in a separate step from this fix. Merging.

PR #48, opened 2015-10-12
Fix mono sqlite - attempt 2

kf6kjg:
It works, both Linux64 and Win64.  No #if tests.


jimtarber:
This will need to be reviewed by Dave when he returns later in the week.

ddaeschler:
Ooh very nice! I'll review ASAP

kf6kjg:
The two biggies I can think of:
* The Linux .so is version 3.8.6 while the Windows .dll and the .NET assembly are version 3.11.1.  Doesn't seem to bother it that the .so is a couple of minors behind: it seems likely that Halcyon isn't using the newer features.
* The Windows .dll is now 64bit whereas it was 32bit before.  If SQLite was written correctly that shouldn't cause any issues when loading pre-existing state files. However if it wasn't, the difference in size_t, void*, &c. might be a problem.

mdickson:
Probably what needs to happen is the DLL needs to be rebuilt on the Windows side with the right options set so it makes both the portable managed DLL and the native SQlite.Interop.dll.  Then the Interop DLL needs to get put into the x86/x64 directory (or a .dll.config set up).  So on Windows it finds things.  On Linux I'd probably use the .so from the OS packaging along with a compiled sqlite3.dll.  Same trick for the Mac with it using the native dylib.

kf6kjg:
I specifically made it not use the interop process following https://gnuforensics.wordpress.com/2013/09/26/how-to-compile-sqlite-system-data-sqlite-on-linux-mono-using-xbuild/ - however I didn't test to see if it could be made to work with interop, so I guess that still might be an option.  However if there's a problem getting the 64bit sqlite3.dll to read state files that were created with a 32bit sqlite3.dll, going interop will cause even more issues - plus still requiring the segregation of linux (64bit only) and Windows 32bit region servers in separate grids.  Again, this is all predicated on IF there's a problem reading a 32bit-made sqlite file with a 64bit sqlite3.dll.

Another issue I ran across: it seems .NET/Mono doesn't know how to go looking for system libraries - nothing worked until I copied the .so from my system library folder to the Halcyon/bin.

A side question: do you have any 32bit region servers?  If not, then it's probably not a problem to just go 64bit.

mdickson:
Well the downside. as I understand it, is that not using Interop means all managed code.  So on Windows it wont be using the optimized Native binaries on Windows.   I may have that wrong, the documentation is  pretty sparse in this area.    Its great you got the managed approach to work though.  If thats good enough then it should ok as is.

kf6kjg:
Well, a proof would be to attempt to run this in Windows without the
sqlite3.dll.  It should fail.

What I did here was to compile on Linux using Mono an edition of
System.Data.SQLite.dll assembly that was designed to connect to a
non-interop native library in a way that worked not matter whether it was a
.so on Linux or a .dll on Windows. I did that following the above linked
GNUForensics instructions.

After that I found that it was complaining that the native library was
missing, so I cp'd from my Ubuntu 64bit system's installed sqlite, found
via the find command somewhere off in in a system 64bit lib folder, into
the Halcyon/bin folder.  I did this assuming that the .NET/Mono assembly
compiled above might be not looking at the system library folders.  Sure
enough I was right: it worked, so I added the .so and the changed assembly
to the repository and committed.

Next step was to test under Windows.  I pulled the changes from my repo,
compiled Halcyon, and found that it threw an exception when trying to
access the sqlite3.dll - wrong bitness.  Sure enough: the sqlite3.dll that
was being shipped with Halcyon was 32bit.  So I pulled the SQLite
amalgamation source and recompiled for 64bit - which was an adventure as
the build instructions on SQLite's site are wrong: follow the instructions
and your dll won't have any exported functions.  Thankfully I found a page
that gave the right instructions.  Got that in place and it worked.  (I'm
skipping over the big long fight trying to comprehend the
not-well-explained exceptions .NET was throwing so I could even find a
correct solution...)

ddaeschler:
I'll review this one tomorrow

ddaeschler:
I've confirmed that under windows the state saves are working, and indeed, when I renamed the native sqlite dll I got an exception as expected:

Exception thrown: 'System.DllNotFoundException' in System.Data.SQLite.dll

I'd have to conclude these changes are using the native sqlite on windows. This begs the question what our previous configuration was doing since the 32 bit DLL should not have loaded in the halcyon process which has 64 bit specified.

In either case, though this is a pretty big change, I'd like to move forward with it. As part of our first Halcyon test rollouht, we just need to make sure this DLL has no problems loading script states in our main grid tests starting with the beta grid and moving on to oceans and mainland.

kf6kjg:
A file I missed, but not critical: `bin/sqlite3.dll.version.txt' with the contents `3.8.11.1` - the idea was to be able to have a note specifying what version the dll was compiled against so that if we need a newer version it's obvious where that one came from.

ddaeschler:
Please add the file in a PR and I'll pull it in

Issue #51, opened 2015-10-14
llGetEnv() sim_channel and sim_version compatibility


kf6kjg:
`sim_channel` and `sim_version` are two requests to llGetEnv that are part of the SL "spec" - they have examples of structure that result in scripter expectations and consequent unforseen uses - but that are not currently matching that defacto spec terribly well.

Firstly: `sim_channel` to me, based on LL's and Inworldz' usage, is the _software name_ - in this case "Halcyon Server" and should be changed to such. Yes that's an incompatible change from InWorldz where the string was "InWorldz Server": the product's name has changed and it cannot be called by the old name anymore, at least not without causing confusion on other grids.  It is not, as some have thought, ever changed across grids: in SL it is "Second Life Server" no matter whether the main grid or the beta grid.  The only place that I can think it would change is if the region was running one of the testing server editions, such as "Magnum". In this case it still is the product name, just a special and highly volatile product name.

Following the above logic, `sim_version` has a problem: it doesn't follow the defacto "spec" in the format of the version string. Following the LL format makes the most sense to support script compatibility for scripts coming from that grid, but InWorlds historically implemented a different format. My suggestion is to revert it to the LL-style version numbering so that scripters fresh from SL get back what they expect: `"11.22.33.9876"` as documented on the LSL wiki instead of `"InWorldz 11.22.33 R9876"` or `"Halcyon 11.22.33 R9876"` as the code currently returns.  Again, the reasoning is that some scripts may be, for whatever reason, parsing the contents and attempting to use the results in unpredictable ways.  As a side note, why is this being so redundant and putting in the server product name?!

Now this means that existing scripts that are attempting to parse these values may have issues: people may have already coded relying on the existing format.  I can't say one way or the other.  What I can see is that when I look at the SL wiki page for `llGetEnv()` I see a string format that will be returned, and when I test in InWorldz or on Halcyon I don't get what I expect based on what I read there.  Another side note: a simple parse of the `sim_channel` would quickly let any script know what simulator type it's on.

I have an incoming pull request waiting on the results of this discussion, however I figured this kind of change needed more input up front before locking in the changes and submitting them for review.

ddaeschler:
See the comments made to #56 as well

jimtarber:
I took a look at llGetEnv results on both InWorldz, Halcyon and SL servers (see output below).

[...]

Issue #54, opened 2015-10-16
What's WhisperVoice?


kf6kjg:
The code is somehow loading it, and is (because of some changes I'm working on) no longer running on my local copy because WhisperVoice isn't able to be loaded for some reason.  My question is: what is/was it, and is it needed any more?

jimtarber:
Whisper is an alternative to Vivox for in-world voice. See Mumble/Murmur https://en.wikipedia.org/wiki/Mumble_%28software%29 

We don't use it in InWorldz (we use Vivox) and I'm not sure what state the Whisper code is in. The last update was December 2011. 

SL uses Vivox, so providing Vivox-based in-world voice also added more compatibility with SL.

In January 2012, we adopted Vivox instead, after we negotiated a license. Also at that time, Vivox began offering free service for small OpenSim grids: https://support.vivox.com/opensim/

The Halcyon project has a philosophy of removing redundant options for various subsystem implementations and going with the best available solutions for each component. So as far as I know, at least for the foreseeable future, we have no intention of supporting Vivox. I believe it's still there because some work was done initially to support Whisper and it's just saving that effort.

kf6kjg:
I'm sure "we have no intention of supporting Vivox." has a typo. :P

Ok, so removing those files wasn't an issue.  Not sure where they came from in my working copy, but the current tree doesn't seem to have them or any of their supporting files anyway.

Thanks!

PR #56, opened 2015-10-16
Cleanup

kf6kjg:
This is a series of patches designed to just remove some of the dirt in the code.  Some things are simple but broad, such as whitespace cleanup and spelling changes or dead code purging, some are more helpful and specific.

In one case however there's a scripting backwards compatibility change: see the comments on 37bcad1 and Issue #51 for more information.

Yes the changes are very big.


ddaeschler:
Wowsers. Give me some time to review this one  :)

ddaeschler:
We're provisionally merging this.

The last commit contains some stuff that could potentially break scripts checking for an InWorldz environment. We're going to need to be able to override that for the IW grid, so @jimtarber @kf6kjg and myself should have a discussion about where best to put the override for backwards compat.

kf6kjg:
Wow, that was faster than I thought! :D

WRT to overriding and how best to accomplish that in a general context, I've created #57.  For how we want to proceed on this specific instance, #51 is what I made for the task!

jimtarber:
I'm slowly working through this one too. There's a lot to cover in this commit.  Overall I'm pleased; it's nice to see this kind of cleanup of old stuff and lesser source code issues (especially things like hard tabs, etc).  But some of the changes are fairly sweeping, and I do have some concerns.

First, I do want to point out that "Initialise" and "Normalise" are actually spelled correctly in England, Canada and just about everywhere outside of the USA. Just changing that one spelling ("Initialize") to a different spelling resulted in most of the size of this commit, which seems to me to be mostly unnecessary work. Please consider the cost/benefit potential of not fixing spelling errors in *interface* names going forward.  (That said, I prefer the 'z' versions myself for these, even though I'm in Canada, and it's kind of the *InWorldz namesake* spelling anyway. At InWorldz, we're kinda known for changing 's' to 'z' in overzealous ways.) :stuck_out_tongue_winking_eye:  And I think the spelling fix for AuthenticateAsSystemUser is a good one, so a bit of a balancing act there depending on impact.

The change to the DATA_SIM_RELEASE option in llRequestSimulatorData is very similar to the llGetEnv("InWorldz") issue.  Again, some software may be using that to distinguish grids, especially in the absence of a "gridnick" option.  I don't think we can just change it to return the software name, at least on InWorldz grids.

Similarly, if the user-agent string in HTTP requests (elsewhere0 is changed from InWorldz to Halcyon, the string comparison in llSetContentType may not be able to detect calls from Halcyon scripts (see the hard-coded test in llSetContentType, not changed by the commit).

I'm also a bit concerned about removing huge chunks of commented-out code (inside #if false tests). That method is sometimes used to put a cap on work in progress, to allow it to be stashed in source control, not lost in a branch, but to ensure that the untested partial code has no effect on the production servers.  If you were the one to comment it out, of course that's different. But a *lot* of code has been removed from RemoteAdmin.cs, which I believe @mdickson was working on some time ago before having to switch gears to materials and other priorities. Similarly for large blocks of disabled code in AssetCapsModule.cs.  Both may also just be very old OpenSim code that we don't use, and can come out, but I think it's important to know for sure that it's not a work in progress before those kinds of significant deletes.

Also sometimes code is disabled but left in place to document a conscious decision to not include it. This helps to avoid someone who knows it's missing from coming along and adding it back (this time not commented out).  A good example of that is the disabled support for estate_id and parent_estate_id in EstateSettings.cs (handleIncomingConfiguration).  We're specifically disabling those two incoming fields so that the viewer user can't just change the estate IDs to something they make up.  Removing that commented-out code puts that operation at risk of being "fixed" down the road. (It also hides the well-known implementation while it is present there but disabled, so that if we ever decide that we *do* in fact want to do something with it, we have to dig to even find out what the configuration key names are that we're intentionally ignoring.  The calculation of SimWideMaxPrims in SendLandProperties is another example of that, where we chose to put an explicit prim count in the region.xml file, but in the absence of that, this disabled code is how it would be calculated (based on land). If we ever decided to support var regions or anything like that, we might need that (disabled) code.

jimtarber:
Still reviewing the changes, but did you check to see that the caller of RESTInterregionComms.cs DoAgentDelete() won't gag on the modified "str_response_string" that now has "Halcyon" in the text? That's a protocol change, and it's probably safe (the caller is probably checking the status code if anything, but it is a change to the protocol that needs to be confirmed to be safe.

jimtarber:
I tried testing an updated master, and I was having an exception at startup loading the modules. After a short investigation, it seems I had an old *InWorldz.WhisperVoice.dll* in my bin folder, and because it did not have a PostInitialize (it had PostInitialise), it was generating a load exception.  Of course the solution is to delete the DLL, but it highlights the kind of problem that can happen with changes to interfaces, and why they should be made when there is good value to the change. (It did not cost much time; the exception report indicated PostInitialize in one of the exception fields and so I knew what the problem was right away.)  This also was only a developer test problem; our production tools remove all DLLs, EXEs and PDBs on each upgrade, so it wouldn't have been a problem on the main grid either.

jimtarber:
Okay, one more issue in the final commit. GridInfoService.cs changes the constant "platform" value from "OpenSim" to whatever the software name is ("Halcyon Server").

This is the data directly returned to the viewer on a call to get the Grid Info.  I'm not sure in what cases a viewer behaves differently when the platform is "OpenSim", and that may depend on the viewer.  So this is also a protocol change, and probably not a safe change.  To go ahead with that one, we'd probably have to check with the IW3 viewer, as well as IW2, Firestorm, Singularity and Cool VL teams at the very least.

jimtarber:
I've discussed that last one with the IW3 viewer developer and IW3 *tests the URI* returned to see if it contains "inworldz.com". Otherwise, e.g. on dev test servers, it only recognizes "OpenSim" as the platform.  Which would be a problem for this change, except that it also defaults to "OpenSim" for unrecognized platforms.  Also, it is believed that other viewers will also default to "OpenSim" for any unrecognized platform value.

So we may be able to change that to indicate Halcyon software, or InWorldz (or *conditionally* one or the other based on that InWorldz=1 proposed INI option). But it looks like we could probably proceed with this change.

More info: The response from the server includes "gridnick", which is what the viewer calls "grid_login_id" (e.g. "Aditi"), and "gridname", which is what the viewer calls "label" (e.g. "Second Life Beta"), and platform is the underlying software tech (e.g. "secondlife"). So in theory it would be suitable for us to plug "halcyon" in that last one, for all halcyon servers and grids, including InWorldz.  However, it seems that platform is typically a single lowercase word, so I'd like us to drop drop the "Server" from that and return just "halcyon".

jimtarber:
There are also some other keys recognized in the grid info response, such as "administrator" which is the grid owner (company). For both SL and SL beta, the administrator is "Linden Lab". So for InWorldz grids that should be "InWorldz LLC", but for other grids something else. So we probably should add an admin="InWorldz LLC" to the grid info in the INI file, and let other grids and test grids specify their own.

kf6kjg:
Good to see an in-depth analysis!

I do admit that I cannot see all possible consequences: but that's why we have these reviews!  Thank you for the exercise in having to dig through and verify that I did do this correctly and didn't wind up creating unexpected breakage!

"Initialise" and "Normalise": I understand that they are "correct" in other countries than the US.  However, a search of the [Oxford English dictionary in UK English for "initialise" returns "initialize",](http://www.oxforddictionaries.com/definition/english/initialize?q=initialise) and the [Cambridge English Dictionary does likewise](http://dictionary.cambridge.org/dictionary/english/initialize?q=initialise).  A [random spelling site lists the "ize" variants as archaic in Britain, but current in both US and Canadian English](http://www.lukemastin.com/testing/spelling/cgi-bin/database.cgi?action=view_category&database=spelling&category=I).  Our own [wiki](https://github.com/InWorldz/halcyon/wiki/Contributing), admittedly under my editing, states to use "ize", however that was based on a review of the InWorldz-added modules which were using "ize" except when interfacing to the OpenSim modules. (Yeah, I figured I'd have to defend this one at some point! :D I didn't hunt down "Normalize", but I presume it's in a similar state.)

DATA_SIM_RELEASE is not LL standard, and it also changed what it returned depending on the test `simulator == World.RegionInfo.RegionName` - which I understand as "is the requested region name the same as the current region?" If that test is true, the returned value was ["OpenSimulator Server"](https://github.com/InWorldz/halcyon/commit/37bcad1321651848c2cab9b2451cf9d7bb5d2acf#diff-13) and if false was ["InWorldz"](https://github.com/InWorldz/halcyon/commit/37bcad1321651848c2cab9b2451cf9d7bb5d2acf#diff-0).  At least it's now consistent as both routes now return whatever's in VersionInfo.FullVersion.

If I read the code correctly, [`llSetContentType`](https://github.com/InWorldz/halcyon/blob/master/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs#L12561) is reading the value returned by the viewer, not a server-server script call. Note the preceding test checks to see if the owner is online and in the same region, and that the comments are about browsers, not scripts.  My scripting experience also indicates that this command is irrelevant to scripts as they don't differentiate on content type: text/plain is fine to them.

As regards the `#if false` commented code, I understand, hence why well-documented blocks were left behind.  If indeed some of this was WIP by @mdickson or code that was not wanted to return, then it would be good to have them back, but this time with some comments indicating why they exist in that state.  Maybe even better in the latter case, just a comment explaining that the code has been removed and why.

The only places in the code that read `"str_response_string"` that are not logging info is [BaseHttpServer.cs](https://github.com/InWorldz/halcyon/blob/master/OpenSim/Framework/Servers/HttpServer/BaseHttpServer.cs#L1388) and that method is deprecated at the [IHttpServer.cs](https://github.com/InWorldz/halcyon/blob/master/OpenSim/Framework/Servers/HttpServer/Interfaces/IHttpServer.cs#L166) level.

WRT to the InWorldz.WhisperVoice.dll problem, I had that myself - see #54.

In regards to the last two comments on further changes, we may need to look into that.  I'm likewise recalcitrant to add new stanzas to the INI unless these things are thought out thoroughly.  Such may need its own issue report to provide a place for discussion.

jimtarber:
Yes, all of this is expected on such a large set of changes and exactly why we have a few eyes reviewing it. Dave reviews all of my changes as well, and I tend to push most commits to a branch and wait for him to review and merge my PRs.  (I tend to review many of his as well, although they tend to be more tactical changes, or new code, and thus either required changes or not likely to have any impact.) It's good to have more eyes on something.

And I agree many of these things are related and need a more unified strategy and discussion. We're busy with the first production use of Halcyon (crossing problems at the moment), so we'll need to defer that temporarily, but I'd like to get the version/conditionals/etc wrapped up asap to enable further work.

jimtarber:
This is already long since merged into master, but just to put a bow on this one, I think #51 and #57 and PR #74 take care of this and allow all of the changes in this pull to roll out to the main grid now.

PR #62, opened 2015-10-21
Add an unloaded script cache to the Phlox loader

ddaeschler:
This will prevent disk thrashing when scripts are rapidly loaded and unloaded during rez/derez of eg temporary objects.


jimtarber:
This is a really great idea, especially for those scripts always re-rezzing objects, since the refcount can go to zero briefly.

PR #64, opened 2015-10-22
Add a DateTime field to CacheEntry and code to Cache to use it and a …

mdickson:
Pull request for initial aged cache implementation.  There are 2 constructors so other than locking the current implementation will work as it already does.  The client code needs to be updated to specify a timer interval and max age to enable the aging behavior.

Room to optimize here but this is functional including a test.


ddaeschler:
Holding this for now pending further discussion

PR #65, opened 2015-10-27
Fixed forced group membership db queries to be optimized if possible.

jimtarber:
Uses the LLCV groups list if there is one. Note: it looks like more
changes since I also fixed the agentID/agentId naming inconsistency so I
could follow it through in the debugger.

The key change is to now call FastConfirmGroupMember instead of
IsAgentInGroup in the groups module (which is a db call).


ddaeschler:
Looks good. I don't see any obvious issues.

PR #66, opened 2015-10-27
Made the object2 post message aware of seated avatar (IDs).

jimtarber:
This is part 1 of a set of fixes and performance improvements to come related to parcel/region checks and user profile caching.

This allows additional crossing pre-checks previously only done on the
object owner UUID.  Does not introduce new cases, just more of the same.
(Used to be able to gain region access by sitting on someone else's
prim, or as passenger on someone else's vehicle. But after crossing,
error handling was horrible.)  Also consolodated region and parcel
checks, removing duplicate specific code (that was sometimes
different/incomplete).


jimtarber:
Actually I see conflicts on this so ignore the PR, I will fix them in the branch first.

PR #67, opened 2015-10-27
Add aging support as well as maintaining a reserve size to the LRU Cache module

mdickson:
Default behavior is no change.  Additional constructors allow enabling the new behaviour.  Added tests as well.


ddaeschler:
Looks good!

PR #68, opened 2015-10-29
Add a default value of 0 for iz_level in the users table so create fr…

mdickson:
…om the console works correctly.


ddaeschler:
Good to go

PR #71, opened 2015-11-03
Web profiles

kf6kjg:
At this point I've tested against FS, and have started documenting the call and developing an example profiles nodejs server.

I still need to finish testing against IW3 and Alchemy.  If both those work then this should be fine.     Just have run shy on time the past couple of weeks. :(


ddaeschler:
I'll merge this after testing with iw3

ddaeschler:
On a standalone simulator using iwv3 I opened up the webprofile and only got an about:blank page instead of the oops page I would expect for that demo URL. Perhaps your last change made it so standalone setups could not use the webprofiles?

Right now, a user server has some of its settings in the xml and some in halcyon.ini. This needs to be cleaned up and I'd love to remove the xml file, but in the meantime you'll probably need to make sure that standalone is also able to get the webprofile values from the halcyon.ini. I'm not sure that is the issue here. Just guessing based on your last commit message.

Please check standalone and let me know.

OpenSim/Framework/Communications/Services/LoginResponse.cs
```diff
@@ -400,6 +401,8 @@ public virtual XmlRpcResponse ToXmlRpcResponse()
                 responseData["region_y"] = (Int32)(RegionY * Constants.RegionSize);
 
                 responseData["map-server-url"] = MapServerURI;
+                responseData["profile-server-url"] = ProfileServerURI;
```
jimtarber: This PR looks good except for these assignments, which should not be made if the string is empty or null, otherwise it triggers the viewer to use them.

jimtarber:
There are conflicts on the pending merge; can you update from master and resolve them?

kf6kjg:
Ok, so I was unable to test the standalone setup as I seem to be in a strange situation: I can set up and connect under gridmode, but can't seem to figure out standalone! :P

Theoretically standalone should work, all the code looks like it's in place.

ddaeschler:
Can you merge in changes from master to your branch here and resolve the conflicts? I should be able to pull in this PR then.

kf6kjg:
So done. :)

PR #72, opened 2015-11-05
Jim/profilecaching

jimtarber:
A fairly significant reworking of user profile caching.  All caching is now in UserProfileManager, with XMLRPC and db queries from UserProfileManagerData. Reworked all the locking and cache structures to use only two locks, one for "level 1" caching ("Data" objects) which is UserProfileData and UserAgentData objects, and one for "level 2" CachedUserInfo objects ("Info").  Caching within OSG1UserDataPlugin has been removed, and the interface to UserProfileManager made more consistent and new lighter methods added to convert Key2Name and Name2Key, which was the most common operation and previously heavier.  Several mismatch bugs with regards to which locks were used previously have been fixed, and now profile updates and forced refreshes no longer clear the cache then force a refresh, but leave it in place until the new data is available, then replace the profile with the update. This means the old data is still available while the new data is fetched, which may be important for Name2Key and Key2Name conversions, and others such as group role abilities.  As soon as the I/O completes, the profile will be updated with the fresh data. The refresh on crossings is now done asynchronously, and the groups checks use in-memory data only, so at least one forced profile update (XMLRPC call) and one complex db query to fetch group info has been eliminated from crossings.  There are no known I/O calls during crossings now.


jimtarber:
Canceling the pull request; still testing, login User server now, and finding issues still. Fixing.

jimtarber:
Okay, the last commit to ensure there's a UserAgentData for any UserProfileData fixes the User server problem where the usage was a bit different than on regions. Ready for a pull.

ddaeschler:
This was really tough to review due to the number of changes. We'll need extra testing love.

PR #76, opened 2015-11-09
Replaced object sale prechecks with warning messages.

jimtarber:
Fixes problems with IW3 and other viewers that don't have an Apply
button in the Edit window, that prevent checking For Sale checkbox for
no-copy or no-trans objects to reach valid object settings.

This addresses the problem seen in http://inworldz.com/forums/viewtopic.php?f=97&t=21794 as well as improves the user interface for all viewers (including IW2).


ddaeschler:
Looks good and should prove very useful

PR #77, opened 2015-11-12
Jim/regionaccess2 -  Crossing improvements and restricted parcel access fixes

jimtarber:
Fairly major changes, first to make the object2 post message aware of seated avatar (IDs) for crossing pre-checks.  Avatars and objects are not allowed to begin a crossing if the destination parcel is not accessible to the object owner and all seated avatars.

Also, within a regions, parcel border crossings are now recognized, and for banned avatars, owners of objects, and passengers, both avatars and non-physical objects are just stopped dead at the parcel border.  For physical objects, we cannot prevent entry so they are instead returned to the owner after unseating the riders at the parcel border.  This fixes lots of problems where the avatars were allowed in if they pushed hard enough, physical objects (including vehicles) were always allowed in, and sometimes only the objects were tested rather than objects and their passengers.

Also some problems with using the seated avatar position (near 0,0,0) as an absolute position, and cases where the wrong location was used to do parcel checks.

### DIFF
--- a/OpenSim/Region/Framework/Scenes/ScenePresence.cs
+++ b/OpenSim/Region/Framework/Scenes/ScenePresence.cs
@@ -523,13 +523,26 @@ namespace OpenSim.Region.Framework.Scenes
                 if ((!forced) && (IsInTransit) && (parent != m_posInfo.Parent))
                     return;
 
+                if (parent == null)
+                {
+                    // not seated
+                    ILandObject parcel = Scene.LandChannel.GetLandObject(agentPos.X, agentPos.Y);
+                    if (parcel != null)
+                    {
+                        ParcelPropertiesStatus reason;
+                        if (agentPos.Z < Scene.LandChannel.GetBanHeight() && parcel.DenyParcelAccess(this.UUID, out reason))
+                        {
+                            if (!forced)
+                                return; // illegal position
+                        } else
+                            lastKnownAllowedPosition = agentPos;
+                    }
+                }
+
                 m_posInfo.Set(agentPos, parent, parentPos);
                 oldVelocity = m_velocity;
                 m_velocity = velocity;
                 ForceAgentPositionInRegion();
-                if (m_posInfo.Parent != null)
-                    if ((m_posInfo.Position.X >= 6.0f) || (m_posInfo.Position.Y >= 6.0f))
-                        m_log.Error("[SCENE PRESENCE]: SetAgentPositionInfo - unexpected position " + m_posInfo.Position.ToString());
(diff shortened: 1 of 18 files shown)


OpenSim/Region/Framework/Scenes/ScenePresence.cs
```diff
@@ -564,13 +577,52 @@ public override Vector3 AbsolutePosition
...
                             Util.ForceValidRegionXY(ref ppos);
+                            m_physicsActor.Velocity = Vector3.Zero;
                             posForced = true;
                         }
+
+                        ILandObject parcel = Scene.LandChannel.GetLandObject(ppos.X, ppos.Y);
+                        if (parcel != null)
+                        {
+                            ParcelPropertiesStatus reason;
+                            if ((ppos.Z < Scene.LandChannel.GetBanHeight()) && (parcel.DenyParcelAccess(this.UUID, out reason)))
```
ddaeschler: Since this block is called by physics updating position info 64x/sec, have we checked to make sure parcel.DenyParcelAccess isn't going to the database etc?

ddaeschler:
The above comment highlights a future enhancement. Right now, there is a mechanism in place to catch when an object crosses into a new parcel, and fire off an event. Having the same mechanism for avatars would allow this code to be run only when they cross into a new parcel, and not for every frame. This may be a moot point if these checks are very light, but something to consider.

jimtarber:
The new calls do not invoke any disk ops, XMLRPC calls or db queries. (That was part of the earlier fixes in the #72 where it used to be a problem in the previous release.)

The current mechanism to catch parcel crossings is significantly flawed (it's why you could cross if you pushed on it hard enough, both in the case of avatars and in the case of prims).  Any mechanism to recognize *and prevent* a crossing due to access restrictions (or for any other reason) must prevent the movement, not catch it asynchronously after that fact.  Both in the case of region borders and parcel borders, dealing with it after already allowing it is how the server gets into serious trouble.

I had planned to eventually take a look at whether the existing mechanism could be removed, but that's not a requirement for this set of changes, because it is currently a partitioning of operations that must be done **at** crossing time versus operations that can be done afterwards (asynchronously).

Also note that the code that is in-line with the position change is only the check for whether to allow the position change or not. Once recognized, the handling of it is performed by a FireAndForget block which must be async due to the various callers of the position change (can easily trigger locking deadlocks otherwise).

If the suggestion here is to tie *that* FireAndForget block to the parcel change handler (which still runs inline in the same thread), I don't really think that's appropriate because it is not parcel change handling code, it is parcel change *prevention* code.

PR #78, opened 2015-11-15
Release 20151112 - Deadlock and local/temp profile fix.

jimtarber:
Fixed the deadlock reported on the sandboxes (hopefully): Split AbsolutePosition, which involves a leaf lock (m_posInfo) into a separate function that sometimes does and sometimes does not do the extra work involving parcel checks and updates from physics.  Only CheckForSignificantMovement does the full version, and it's safe to call it from that.  This is also a performance optimization to simplify the code for all of the other calls to AbsolutePosition.
Also fixed the handling of forced refreshes for local and temp (bot) profiles. It was forcing data fetches for local users, and failing to recognize existing temp (bot) profiles.

### DIFF
--- a/OpenSim/Framework/Communications/UserProfileManager.cs
+++ b/OpenSim/Framework/Communications/UserProfileManager.cs
@@ -233,12 +233,23 @@ namespace OpenSim.Framework.Communications
             if (uuid == UUID.Zero)
                 return null;    // fast exit for no user specified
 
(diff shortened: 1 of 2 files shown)


jimtarber:
Would be good to merge this release branch back into master, but keep the branch until after the RC is rolled out.

ddaeschler:
Merged and kept.

PR #79, opened 2015-11-15
Abs pos and misc

jimtarber:
Avoid calling the AbsolutePosition recalculation repeatedly just to get .X, .Y, .Z ... It can be an expensive operation to just waste like that. Also factored LocationURL to Util.cs, useful to other code, and also avoids calling the AbsolutePosition calculation three times. Also fixed EntityBase to lock the posInfo before changing or returning it.


jimtarber:
Note: These changes are NOT in the release branch. They are for the next release but aren't in master yet.

PR #80, opened 2015-11-21
dev/mdickson/materials region support

mdickson:
Two additional points.  You need to add a new field to the shape table.  

ALTER TABLE primshapes ADD Materials longblob AFTER Media;

I'll work on schema evolution which we need in general as a separate commit.

Also Add the following to Halcyon.ini

[RenderMaterials]
	Enabled = true
	DebugEnabled = false

Both are false by default.  Enabled=true enables the RenderMaterials CAPS.  DebugEnabled=true turns on debug logging. The default is to log at the Warn level.


InWorldz/InWorldz.Region.Data.Thoosa/Serialization/PrimShapeSnapshot.cs
```diff
@@ -214,6 +214,9 @@ public byte[] SculptData
         [ProtoMember(55)]
         public int LowestLODBytes;
 
+        [ProtoMember(56)]
+        public byte[] RenderMaterials;
+
```
ddaeschler: Since RenderMaterials is already a protocontract, it would be more efficient to just use that class here rather than serializing first

OpenSim/Framework/PrimitiveBaseShape.cs
```diff
@@ -1051,6 +1071,13 @@ public ulong GetMeshKey(Vector3 size, float lod)
                 hash = djb2(hash, this.SculptType);
             }
 
+            if (this.RenderMaterials != null)
```
ddaeschler: I dont think anything about render materials affect the physics shape. If this is the case, the materials should not be included in the hash since two objects with different materials would still have the same shape.

PR #81, opened 2015-11-25
Significant performance and reliability fixes in RC.

jimtarber:
Fixed parcel access checks to be much less frequent and lighter, and fixed uncaught MySQL connection exceptions:
- Added exception handling wrapper to GetConnection (exception seen on Beta), and updated all GetConnection calls to check for a null return and handle as appropriate for the function. Also updated console message calls to be consistent with other modules (m_log and "[GROUPS]:" with a colon.
- Optimization for IsAgentInGroup to check the in-memory LLCV data first. May avoid db queries at critical times.
- Fixes and performance optimizations to isRestrictedFromLand(), which now checks Public Access first, and does not do group checks unless the parcel settings mean they matter to prim/user access. Also fixes some problems with inverted logic, e.g. group access test applying when Public access was enabled.  Also added some easier/refactored variants of FastConfirmGroupMember.

### DIFF
--- a/OpenSim/Region/CoreModules/World/Land/LandObject.cs
+++ b/OpenSim/Region/CoreModules/World/Land/LandObject.cs
@@ -493,59 +493,44 @@ namespace OpenSim.Region.CoreModules.World.Land
             if (m_scene.Permissions.BypassPermissions())
                 return false;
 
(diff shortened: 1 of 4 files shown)


jimtarber:
Request to merge but NOT delete the branch.

PR #87, opened 2015-12-03
Enable aging of the bufferpool

ddaeschler:
This is the final step to enable increasing the asset cache sizes. The asset cache is bufferpool backed which meant we needed to prune out entries that haven't been used in a while.

Also includes a change to LRUCache to drive the maintenance of LRU from outside of the class so we don't have to make it IDisposable.


jimtarber:
Passes my review.

jimtarber:
Merged.

PR #89, opened 2015-12-06
Projection LSL Support

kf6kjg:
Corresponds to Phlox PR#6 and already includes a rebuilt edition of the Phlox DLL with the new constants and rules that enable script control of the projectors.

Please use Phlox PR#6 for discussing the engineering of the constants.  Eg: whether they should be individual constants like this, or a glob set like most of the commands - or if you'd like to see the naming be different.

The re-engineering of `KeyOrName` I'm no longer certain is 100% needed, as such is not an exploit, but it does make things act more consistently: inventory items, no matter the name, are now accessed as such - even if the __name__ of the inventory item is "6522E74D-1660-4E7F-B601-6F48C1659A77".


kf6kjg:
Done.

kf6kjg:
And the bulk command is done, making this feature complete methinks!

jimtarber:
Looks good, merged.

PR #94, opened 2015-12-08
Jim positionchecks

jimtarber:
Also needs to recognize decreasing Z values (entering a parcel from above) but not increasing values.


jimtarber:
Retested, passed.

PR #97, opened 2015-12-09
Plumb the message generator avatar ID through chat

ddaeschler:
GeneratingAvatarID added to OSChatMessage to track the avatar that is responsible for a message regardless of of it comes from one of their objects, or the avatar itself.

Removed the field from OSChatMessage that was storing a reference to the SOP that was generating messages in the case of scripted chat. This was only being used to look up the avatar, and doing it badly.


jimtarber:
This looks good to me. (And hugely improved over that very ugly old code.)

PR #99, opened 2015-12-13
Inline code documentation & Doxygen LSL documentation starter

kf6kjg:
Comments from PR 95, now part of this:
Part of a probable series of code documentation work that gets done as it annoys me and isn't otherwise part of actual work on the code.

Some of the changes seem odd from this bulk change view, so be sure to dig into the commit messages: lots of explanations in there.

Comments for this PR:
This allows Doxygen to build LSL documentation from inline docs via specific-purpose pages.

I'd experimented with using Doxygen's ALIASES, but found that it didn't want to work for some reason.  I left them in in case they can be made to work: I suspect the issue was simply a doxygen version issue on my end.

The goal would be to have something that makes it easy to build wiki pages and more from - or even to publish directly if it seems reasonable.

Examples: 
* The pages list: The pages list
* The LSL Constants page The LSL Constants page
* The ScriptBaseClass Class Reference page The ScriptBaseClass Class Reference page


jimtarber:
"This allows Doxygen to build LSL documentation from inline docs via specific-purpose pages."

What is the goal here? There is already an InWorldz LSL wiki, with documentation of IW-specific LSL information.  For example, here are the [IW-specific LSL functions] (http://wiki.inworldz.com/izwiki/index.php/Category:IWSL_Functions) and an example, [iwGetAgentList] (http://wiki.inworldz.com/izwiki/index.php/IwGetAgentList). It would be nice to have some of this auto-generated from source, however that is a huge project to retrofit into a large existing code base.

I like this, maybe even love it, but I'm not sure it's very practical in a large existing project, especially one that already has its own wiki that volunteers have been pouring effort into for years.

kf6kjg:
The purpose is to bring baseline documentation closer to the code.  It's not a replacement for the wiki by any means, it's a way that someone who's already elbow deep in the code can add baseline documentation while they are already there.

By baseline I mean that it's information that can be used to build a wiki entry.  It also not as subject to editor's whim: it's docs added by the developer explaining the intent of the item.

While the developer could be so bold as to go create the wiki pages directly, it's another external step.

mdickson:
Halcyon is an open source project and having tools to maintain documentation attached to the project rather than in just an external wiki is I think totally appropriate.  And Doxygen is an excellent tool for the job. Might be nice eventually to adapt the wiki to take the Doxygen input.

ddaeschler:
Can you fix the merge conflicts here and I'll review this.

kf6kjg:
Should I just retract PR 95 as it seems that it would be integrated with this?

ddaeschler:
If it is a subset of this one then yes please do

kf6kjg:
Conflicts resolved.

kf6kjg:
There were some bad merges in there WRT the materials addition that happened at the same time as this.  Solved, my re-review of the diff now looks clean to me.

kf6kjg:
In case it was missed, I sync'd with Master again.

InWorldz/InWorldz.PhysxPhysics/PhysxScene.cs
```diff
@@ -278,7 +278,7 @@ public override void Initialize(IMesher meshmerizer, Nini.Config.IConfigSource c
 
             Nini.Config.IConfig startupcfg = config.Configs["Startup"];
             if (startupcfg != null)
-                _gridmode = startupcfg.GetBoolean("gridmode", true);
+                _gridmode = startupcfg.GetBoolean("gridmode", false);
```
kf6kjg: This brings PhysX's default into line with everywhere else in the code that reads the gridmode.

ddaeschler:
Thank you for explaining that change. It was actually the thing holding this up and I hadn't had time to start a dialogue about it.

ddaeschler:
Did you confirm that the changes you made to the sample INI don't affect the defaults chosen for startup?

kf6kjg:
All the ones that I commented out were assigning the default. Connection strings were reformatted to use things like USERNAME instead of "inworldz" and to include the password field.

GuestRegionName was changed to the more generic value "Grid Guests" from "InWorldz Test".

rdb_connection_template was changed to optionally prepend the "Data Source" automatically if it is not already in the string.  Thus existing INI files will go through the original code path.  More information about this particular change is at https://github.com/InWorldz/halcyon/pull/99/commits/b4031c2b83a776e748ed2e335d7022a9e33501b4

OpenSim/Region/CoreModules/ServiceConnectors/Interregion/RESTInterregionComms.cs
```diff
@@ -91,9 +91,7 @@ public virtual void Initialize(Scene scene, IConfigSource config)
 
 
                 IConfig startupConfig = config.Configs["Communications"];
-                if ((startupConfig == null)
-                    || (startupConfig != null)
-                    && (startupConfig.GetString("InterregionComms", "RESTComms") == "RESTComms"))
+                if (startupConfig != null && startupConfig.GetString("InterregionComms", "RESTComms") == "RESTComms")
                 {
```
jimtarber: while the second startupConfig != null test was previously redundant (because of the || ), this commit changes the semantics for if there's no Communications section at all, from defaulting to using this module to not using it.  That seems like it is the improved behavior; I'm just noting the semantic change (which should be okay here -- I don't think we'll have regions that _don't_ have a Communications section and _do_ want RESTComms enabled. As a bit of an aside though, I do wish code like this checking for specific names was case-insensitive.

jimtarber:
This one looks good to me, ready for merge.

OpenSim/Region/CoreModules/ServiceConnectors/Interregion/RESTInterregionComms.cs
kf6kjg: I agree with that aside. Yet another TODO list item :)

kf6kjg:
All the ones that I commented out were assigning the default. Connection strings were reformatted to use things like USERNAME instead of "inworldz" and to include the password field.

GuestRegionName was changed to the more generic value "Grid Guests" from "InWorldz Test".

rdb_connection_template was changed to optionally prepend the "Data Source" automatically if it is not already in the string.  Thus existing INI files will go through the original code path.  More information about this particular change is at https://github.com/InWorldz/halcyon/pull/99/commits/b4031c2b83a776e748ed2e335d7022a9e33501b4
Did I misunderstand the question?

PR #101, opened 2015-12-16
Basicphysics again

kf6kjg:
Agents can fly nicely, stand/walk jerkily on terrain, prims can spin and do other scripted actions. Prims cannot be physical or even be able to be collided against - so no VolumeDetect etc.

The issue with terrain is one of physics engine design: this is so simple that it causes some penetration/rejection oscillation.  If you want your avvie to be still have it sit in a prim or fly.


ddaeschler:
Looks good and not much chance for regressions since most of it is new code. Merging.

PR #105, opened 2015-12-21
Fixing the float to string conversions to match SL.

kf6kjg:
With related change to Phlox, this corrects llDumpList2String but leaves llList2CSV using the short notation.

You can compare the differences between SL and Halcyon/Phlox using the following script - also attached.

``` LSL
test_str(string name, string expected, string value) {
    if (value == expected)
        llOwnerSay("PASSED - " + name);
    else
        llOwnerSay("FAIL - " + name + "\n  Expected '" + (string) expected + "' but got '" + (string) value + "'");
}

default {
    state_entry() {
        llOwnerSay("==== STARTING TESTS ===="); // Tests tuned to pass on Second Life Server 15.12.01.308474

        test_str("Zero integer cast to string", "0", (string)0);
        test_str("Integer cast to string", "1", (string)1);
        test_str("Negative integer cast to string", "-1", (string)-1);
        test_str("High value integer cast to string", "12345", (string)12345);
        test_str("Zero float cast to string", "0.000000", (string)0.0);
        test_str("Integral float cast to string", "1.000000", (string)1.0);
        test_str("Negative integral float cast to string", "-1.000000", (string)-1.0);
        test_str("High value integral float cast to string", "12345.000000", (string)12345.0);
        test_str("PI cast to string", "3.141593", (string)PI);
        {
            list lst = [0.0, 1.0, PI, -1.0];
            test_str("Four floats in a list cast to string", "0.0000001.0000003.141593-1.000000", (string)lst);
        }
        {
            list lst = [0.0, 1.0, PI, -1.0];
            test_str("Four floats in a list dumped to string", "0.000000|1.000000|3.141593|-1.000000", llDumpList2String(lst, "|"));
        }
        {
            list lst = [0.0, 1.0, PI, -1.0];
            test_str("Four floats in a list CSV'd to string", "0.000000, 1.000000, 3.141593, -1.000000", llList2CSV(lst));
        }
        test_str("Vector cast to string", "<0.00000, 1.00000, 3.14159>", (string)<0, 1, PI>);
        test_str("Rotation cast to string", "<0.00000, 1.00000, 3.14159, -1.00000>", (string)<0, 1, PI, -1>);
        {
            list lst = [<0, 1, PI>, <0, 1, PI, -1>];
            test_str("Vector and rotation in a list cast to string", "<0.000000, 1.000000, 3.141593><0.000000, 1.000000, 3.141593, -1.000000>", (string)lst);
        }
        {
            list lst = [<0, 1, PI>, <0, 1, PI, -1>];
            test_str("Vector and rotation in a list dumped to string", "<0.000000, 1.000000, 3.141593>|<0.000000, 1.000000, 3.141593, -1.000000>", llDumpList2String(lst, "|"));
        }
        {
            list lst = [<0, 1, PI>, <0, 1, PI, -1>];
            test_str("Vector and rotation in a list CSV'd to string", "<0.000000, 1.000000, 3.141593>, <0.000000, 1.000000, 3.141593, -1.000000>", llList2CSV(lst));
        }
    }
}
```

With the following commits and the corresponding changes  to Phlox, in Phlox [PR 9](https://github.com/InWorldz/phlox/pull/9), the output is as follows:

```
==== STARTING TESTS ====
[18:32] Primitive: PASSED - Zero integer cast to string
[18:32] Primitive: PASSED - Integer cast to string
[18:32] Primitive: PASSED - Negative integer cast to string
[18:32] Primitive: PASSED - High value integer cast to string
[18:32] Primitive: PASSED - Zero float cast to string
[18:32] Primitive: PASSED - Integral float cast to string
[18:32] Primitive: PASSED - Negative integral float cast to string
[18:32] Primitive: PASSED - High value integral float cast to string
[18:32] Primitive: PASSED - PI cast to string
[18:32] Primitive: PASSED - Four floats in a list cast to string
[18:32] Primitive: PASSED - Four floats in a list dumped to string
[18:32] Primitive: FAIL - Four floats in a list CSV'd to string
  Expected '0.000000, 1.000000, 3.141593, -1.000000' but got '0, 1, 3.141593, -1'
[18:32] Primitive: PASSED - Vector cast to string
[18:32] Primitive: PASSED - Rotation cast to string
[18:32] Primitive: PASSED - Vector and rotation in a list cast to string
[18:32] Primitive: PASSED - Vector and rotation in a list dumped to string
[18:32] Primitive: FAIL - Vector and rotation in a list CSV'd to string
  Expected '<0.000000, 1.000000, 3.141593>, <0.000000, 1.000000, 3.141593, -1.000000>' but got '<0, 1, 3.141593>, <0, 1, 3.141593, -1>'
```

IE: it is still failing on llList2CSV, but unless there's want to do so I'm of the opinion that CSV can stay in the short form.

With this PR is a request to update the included Phlox DLL so that everything lines up.

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
@@ -8581,9 +8581,9 @@ namespace InWorldz.Phlox.Engine
                 return String.Empty;
             }
             StringBuilder ret = new StringBuilder();
-            foreach (object o in src.Data)
+            for (int index = 0; index < src.Data.Length; ++index)
             {
-                ret.Append(o.ToString());
+                ret.Append(src.GetLSLStringItem(index));
                 ret.Append(seperator);
             }
             ret.Length -= seperator.Length;


kf6kjg:
In case anyone is wondering how Phlox currently responds to this script, here are my results with Halcyon 0.9.26:
```
==== STARTING TESTS ====
[22:01] Primitive: PASSED - Zero integer cast to string
[22:01] Primitive: PASSED - Integer cast to string
[22:01] Primitive: PASSED - Negative integer cast to string
[22:01] Primitive: PASSED - High value integer cast to string
[22:01] Primitive: PASSED - Zero float cast to string
[22:01] Primitive: PASSED - Integral float cast to string
[22:01] Primitive: PASSED - Negative integral float cast to string
[22:01] Primitive: PASSED - High value integral float cast to string
[22:01] Primitive: PASSED - PI cast to string
[22:01] Primitive: PASSED - Four floats in a list cast to string
[22:01] Primitive: FAIL - Four floats in a list dumped to string
  Expected '0.000000|1.000000|3.141593|-1.000000' but got '0|1|3.141593|-1'
[22:01] Primitive: FAIL - Four floats in a list CSV'd to string
  Expected '0.000000, 1.000000, 3.141593, -1.000000' but got '0, 1, 3.141593, -1'
[22:01] Primitive: PASSED - Vector cast to string
[22:01] Primitive: PASSED - Rotation cast to string
[22:01] Primitive: FAIL - Vector and rotation in a list cast to string
  Expected '<0.000000, 1.000000, 3.141593><0.000000, 1.000000, 3.141593, -1.000000>' but got '<0.00000, 1.00000, 3.14159><0.00000, 1.00000, 3.14159, -1.00000>'
[22:01] Primitive: FAIL - Vector and rotation in a list dumped to string
  Expected '<0.000000, 1.000000, 3.141593>|<0.000000, 1.000000, 3.141593, -1.000000>' but got '<0, 1, 3.141593>|<0, 1, 3.141593, -1>'
[22:01] Primitive: FAIL - Vector and rotation in a list CSV'd to string
  Expected '<0.000000, 1.000000, 3.141593>, <0.000000, 1.000000, 3.141593, -1.000000>' but got '<0, 1, 3.141593>, <0, 1, 3.141593, -1>'
```

jimtarber:
I was initially pleased to see this one but I think @ddaeschler has some concerns about how existing scripts might deal with this change. In particular, examples like:
```
[22:01] Primitive: FAIL - Four floats in a list dumped to string
  Expected '0.000000|1.000000|3.141593|-1.000000' but got '0|1|3.141593|-1'
```
Scripts that have been seeing that text start with '0|1|'... might fail to parse the line after an update.  I think this is one of those cases where we might need a more solid reason to change it (since the data is actually correct) than a difference in the output formatting for floats.

Are you aware of any SL scripts that break when they come to IW due to the formatting differences?

kf6kjg:
Any one that tries to determine type from format.  And they do exist :P they are hacky, but they exist.  Most all scripts should be assuming they know the format ahead of time and just typecasting, but some don't.

More of import are scripts that display the values, either in chat or in hover text - a common debugging routine.  Having the format jitter so badly when one instant the text is "0.123456" and the next it is "0" and back again is downright annoying and IIRC is how I found this.  Makes watching values as they change nearly impossible.

mdickson:
I'm with Ricky on this one.   Certainly there are  number of tests in the lsl JSON test suite that fail due to floating point formatting.  And since the format produced in SL is so well substantiated it seems very likely that code does exist that would expect that. Ideally a parser should be flexible, you'd hope thats the case.  IMO the most common case: casting between a string and a float needs to work as expected. after that the rest could be classified as bugs in LSL scripts handling floating point data.

mdickson:
One more thought.  Our current implementation is arguably wrong because it will convert a floating point value like 1.0 to a fixed point value '1' when written as a string. We do handle FP numbers that are not whole and correctly display them with 6 digits of precision.  But someone writing a parser could fail to detect that the value 1 WAS floating point and parse it as an integer.  Still the same value but not the correct type.

jimtarber:
I was initially in favor of the merge as well. I do think we _could_ break some content here, but if true, it will probably either have little effect (can be fixed) or won't matter in the big picture. I think Tranq just had some concerns and perhaps wanted to it discussed a bit. I'm not sure how large his objections are; he will need to chime in here. I'm mostly right on the fence on this, and agree with both viewpoints, although I think I'm still leaning in favor of the merge, but he will need to decide.

ddaeschler:
My goal for discussion was to try to think of cases where this might break content on a massive scale such as settings saving/loading and common uses of that nature. I'm in favor of rolling this in and testing the waters on a rollout. If we wind up with massive problems, we can roll back and create a setting where the new formatting is optional, and the default for new halcyon instances.

mdickson:
So the other question I have is related to how its implemented.  Perhaps it should be done in toString in the OSD Float class so its picked up across the board.

jimtarber:
The commit takes advantage of the float to LSLString constructor which matches the SL format. In my opinion, that is the best place as it is specifically formatting a float as a string for LSL purposes.  It only supports the one single format.
```string s = String.Format(Culture.FormatProvider, "{0:0.000000}", f.value);```
Overriding one of the ToString methods would probably either require adding support for all possible ToString formats, or break any code that passes a format string to ToString. It's also much more difficult to know how far-reaching the effects would be.

The commit here is a very targeted change to ```llDumpList2String``` only, and I think that's the best way in this case. I think if there are other places that might also need to use the ```LSLString(float)``` formatter, that there would be a very small number of those and the code in those places can also be specifically updated to use the ```LSLString(float)``` constructor so as to ensure a specific format, and probably need a separate review like this one.

appurist:
> "The commit here is a very targeted change to llDumpList2String only"

This PR was originally submitted at a time of updates turmoil and so was postponed as an unnecessary risk at the time.  As discussed separately, the team feels it is time to apply this one. I don't have any strong concerns about compatibility and as a bonus this change makes Halcyon more SL-compatible as well. I'm going ahead with the merge now.

PR #106, opened 2015-12-21
Fixed GetUserProfile to never return with a null CurrentAgent.

jimtarber:
The (first,last) variant now calls this uuid-based variant which did not
ensure CurrentAgent was initialized like the name-based one did.

### DIFF
--- a/OpenSim/Framework/Communications/UserProfileManager.cs
+++ b/OpenSim/Framework/Communications/UserProfileManager.cs
@@ -240,6 +240,12 @@ namespace OpenSim.Framework.Communications
                 profile = TryGetUserProfile(uuid, false);
                 if (profile != null)
                 {
+                    if (profile.CurrentAgent == null)
+                    {
+                        // Make sure we also have an AgentData for the profile.
+                        profile.CurrentAgent = GetUserAgent(uuid, true);
+                    }
+
                     // Check if we should force a refresh.
                     if (!forceRefresh)
                         return profile;


jimtarber:
I believe this fixes the second last login problem, where it sends the user to Home problem.  This is because UserProfileManager.cs CreateAgent uses profile.HomeLocation if profile.CurrentAgent is null.

OpenSim/Framework/Communications/UserProfileManager.cs
```diff
@@ -240,6 +240,12 @@ public UserProfileData GetUserProfile(UUID uuid, bool forceRefresh)
                 profile = TryGetUserProfile(uuid, false);
                 if (profile != null)
                 {
+                    if (profile.CurrentAgent == null)
```
kf6kjg: I think you might need to have `&& !(m_localUser.ContainsKey(uuid) || m_tempDataByUUID.ContainsKey(uuid))` in the conditional or you wind up violating the comment below that says "Never force refresh of local profiles or temp profiles."

jimtarber: The AgentData isn't the profile. But even if it was, this fix is not any kind of refresh, the *problem* is that the AgentData is null.  The last logoff location is in that object, and it has never been *loaded* in the first place.  For the case of refreshes, we tie the refresh of that to main profile refreshes, but again this isn't a refresh, it's null.

PR #108, opened 2015-12-22
Add an XML serializable RenderMaterialsBytes

mdickson:
Add an XML serializable RenderMaterialsBytes so we can save/restore OARS with materials set on them.  RenderMaterials is still set to XmlIgnore.  It contains a dictionary which the XML serializer chokes on.


ddaeschler:
Looks good. The unit test failures will be handled separately when we re-do backups.

PR #109, opened 2015-12-22
Implement cache aging for the asset cache

ddaeschler:
This PR adds a timer to the stratus asset cache to perform aging based on the previous aging commits.

### DIFF
--- a/InWorldz/InWorldz.Data.Assets.Stratus/Cache/Cache.cs
+++ b/InWorldz/InWorldz.Data.Assets.Stratus/Cache/Cache.cs
@@ -90,13 +90,19 @@ namespace InWorldz.Data.Assets.Stratus.Cache
             }
         }
 
-        public Cache()
+        /// <summary>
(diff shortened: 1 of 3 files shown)


jimtarber:
This looks good, with one possible concern.  I'm not clear about the effects of creating the _bufferPool with the expiry of maxEntryAge in the new code.  Could that present any race conditions at expiry time?  For example, if it was maxEntryAge+something, that would feel safer to me, but I'm not sure who owns that memory. ByteBufferPool manages its own expiry via the ByteBufferPool idleBufferMaxAge parameter, but the LRUCache code *also* calls _bufferPool.ReturnBytes(). If those two are not fighting with each other, but looks good to you, then this is ready to merge.

ddaeschler:
BufferPool technically doesnt own any memory that is leased. It is removed from the pool entirely until it is returned. As far as I can tell because of this there really shouldn't be any races.

PR #111, opened 2015-12-23
Address an issue with duplicating calling cards ...

jimtarber:
each time the user logs in with some v3 viewers. This commit modifies the handling of the "CreateInventoryItem" packet to check whether the new item is a calling card, whether it is in the "Calling Cards"/Friends/All folder, and that another calling card for the user does not already exist in this folder. If a duplicate item exists, the create inventory item action does not take place and is ignored. To make sure that this does not hit the inventory database too much, only one thread is allowed to populate the calling card cache and the others will wait until it completes or 10 seconds have passed. I'm not sure of a better way to do this, but I don't believe it is a problem; as far as I know, UDP packets are processed on a single thread, so the thread handling code *should* never come into effect.

See https://bitbucket.org/cinderblocks/inworldz/commits/e4698bebe1e6e86cfd77dc296b476f1c28e8799c for the reason the viewer is creating these duplicate calling cards.

One issue that this commit does not address is the requirement to only do this check in the first 10 minutes of the user's login. I was unable to find a property that stored when the user logged in and as we will only ever populate this if they create a calling card, I did not feel that the cache would cause too much of a performance hit.

One change that might need to be made to this commit is disabling the warning message I added to show when blocking of these items occurs. I found this helpful for debugging and it may end up being useful for a short period of time for verification on the main grid itself - but it should be disabled in the long term most likely.

This does not address a separate issue that I ran into where a user can modify the description field of a calling card and change it from the UUID of the person on the calling card to something else, causing the viewer to regenerate the calling card upon logging in. This could also be addressed in this new method by adding a check to make sure the UUID is valid and update the inventory item if it is not.

### DIFF
--- a/OpenSim/Framework/Communications/Cache/CachedUserInfo.cs
+++ b/OpenSim/Framework/Communications/Cache/CachedUserInfo.cs
@@ -75,6 +75,15 @@ namespace OpenSim.Framework.Communications.Cache
         public bool HasReceivedInventory { get { return true; } }
 
         /// <summary>
+        /// This is a list of all of the inventory items that are of type "CallingCard" in the
+        ///   "Calling Card"/Friends/All folder to work around a bug with the viewer creating
+        ///   multiple copies of the same calling card
+        /// </summary>
+        private List<string> _namesOfCallingCardsInCallingCardsFriendsAllFolder = null;
+        private object _namesOfCallingCardsInCallingCardsFriendsAllFolderLock = new object();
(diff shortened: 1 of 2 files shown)


jimtarber:
I don't know why this is showing file diffs. It should be exactly the same changes from ea425eadb414929e3e508ef997ccd02947bd1e72 which was part of #100 and also cherry-picked into the release branch, so there should not be file diffs unless the earlier application of #100 somehow didn't apply to master.

ddaeschler:
I'm not sure what is up with that either. These files are changed in master.. I'm going to apply this and see.

PR #112, opened 2015-12-23
 Fixed the region teleport console messages, so that...

jimtarber:
… they indicate the proper destination region.  Was reporting the current region for out-of-region teleports. Now it indicates the current region, a neighbor region at (x,y), or the name of any remote destination region.  (It doesn't have the neighbor name handy so it indicates the position instead of fetching it, and of course, instead of the wrong stinkin' original local region name.)


ddaeschler:
Looks good

PR #114, opened 2015-12-24
Lockfree group parts

ddaeschler:
This is a complete rewrite of the scene object group part storage using immutable data structures to provide lock free reads. Reads constitute the majority of accesses to SOG part internals, so this removes locking from a large swath of the codebase that has caused numerous deadlocks in the past.

I have provided lock free reads on both the bulk SOP data within groups when only simple iteration of all parts is necessary as well as lock free reads to both the global and local part IDs.

There may be some loss of consistency with the locks removed, but the majority of impact looks very minimal. My biggest worry is around link and unlink which may require a swap of the whole immutable GroupPartsCollection after the various manipulations have fully completed during these operations to yield a consistent state atomically. Heavy testing is required to determine if this will be an actual issue at runtime.


jimtarber:
Reviewed, discussed, looks good as-is.

Issue #113, opened 2015-12-24
Orphaned records in inventoryitems for Current Outfit folder content


Vinhold:
When an avatar creates an outfit, a linked record is created in the Current Outfit folder with assetType=25 and the AssetID is the folderID folder placed in the My Outfits folder in the inventoryfolders table. When the Outfit is removed from the Appearance or the Outfits in the Inventory window, the linked record in the Current Outfit folder is not removed. Over time an accumulation of a lot of orphaned records build up in the inventoryitems table. I recommend that the process called to remove an outfit, also include removing the inventoryitem record where assetID=inventoryitem.folderID and assetType=25.

-- List of orphan records in inventoryitems
Select *
From inventoryitems
Where assetType=25 and assetID not in
 (Select folderID From inventoryfolders)

kf6kjg:
Is this removing the currently worn outfit or removing an outfit that is not currently being worn?

kf6kjg:
Doing detailed testing with an instrumented Halcyon instance, a duplicate is made when the user logs in and does a SaveAs of the currently worn outfit.  For some reason the viewer does not know that the previously worn outfit was being worn and does not send a request to remove it.

I can confirm that upon login the viewer (Firestorm 4.7.5 (47975)) marks the avatar as having "No outfit".

I think this is the real bug here: either the viewer is not handling correctly what the avatar is currently wearing at login or Halcyon isn't sending it right.  Going to test with IW3.

kf6kjg:
Cannot replicate in IW3 3.0.2 - the viewer sees what's being worn without a problem.

However, it still might be prudent to have Halcyon clean up any old links when a new outfit link is created without waiting for the viewer to send the remove command.

ddaeschler:
I think this issue can be closed once the changes to remove the extra links have been tested.

Vinhold:
Just for the record, here is a test I did on several viewers and done in Gospel World, InWorldz and Second Life: http://www.gospellearningcenter.com/resources/viewer%20-%20grid%20tests/default.html
I also did a follow up test and report: 
**************************
Viewer Active Outfit Test:
I used IW V3 to clean and set the active outfit. Ran all test viewers to see who got the active outfit.
Firestorm, Alchemy, Kokua could not detect the Active Outfit, but correctly set all the options and displayed the content in COF. Singularity appeared to have it correct.

Clearing the cache in all viewers had minor change in the results.
Only Alchemy got the correct Active Outfit as a result of cache change.

[...]

PR #116, opened 2015-12-27
Workaround for Issue 113

kf6kjg:
Christmas gift for Vin, and by extension the rest of the Halcyon universe, a cleaner database.

This does NOT solve the underlying issue to #113, it only cleans up and makes sure that the mess making doesn't go anywhere. I do not know why FS is/was not detecting the current outfit on a fresh login and sometimes forgetting which one it's wearing mid login, but this patch solves that by making sure the duplicate records go away.

Please review with a fine-tooth comb and a sense of distrust: this does some pretty heavy-handed activity.

### DIFF
--- a/OpenSim/Region/CoreModules/Avatar/AvatarFactory/AvatarFactoryModule.cs
+++ b/OpenSim/Region/CoreModules/Avatar/AvatarFactory/AvatarFactoryModule.cs
@@ -112,11 +112,13 @@ namespace OpenSim.Region.CoreModules.Avatar.AvatarFactory
 
             private static InventoryFolderBase GetCurrentOutfitFolder(CachedUserInfo userInfo)
             {
-                InventoryFolderBase CurrentOutfitFolder = null;
+                // Duplicate method exists at Scene.Inventory.cs::Scene::GetCurrentOutfitFolder
+
+                InventoryFolderBase currentOutfitFolder = null;
 
                 try
                 {
-                    CurrentOutfitFolder = userInfo.FindFolderForType((int)AssetType.CurrentOutfitFolder);
+                    currentOutfitFolder = userInfo.FindFolderForType((int)AssetType.CurrentOutfitFolder);
                 }
                 catch (InventoryStorageException)
                 {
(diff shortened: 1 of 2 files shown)


kf6kjg:
PS: WRT AvatarFactoryModule.cs::GetCurrentOutfitFolder's case change - since I was adding the comment and causing (eww) duplication, I figured I could clean up the mis-used CaseMode in that method.

kf6kjg:
The reason why this works is rather interesting:
* The Current Outfit Folder is a true folder in the system but it has the special job of holding the current "pointer" to the outfit that's being worn. This can be explored on a MySQL-based setup by executing the following statement with the AGENT_UUID string replaced:
```SQL
SELECT * FROM inventoryfolders WHERE agentID = 'ee1602c0-ae2c-4867-a893-509f6c88ba0b' AND type = 46;
```
* The COF should only contain one set of items for any length of time: a singular entry of AssetType.FolderLink (value 25) that points to the outfit folder, and a collection of AssetType.Link (value 24) that each point to the respective items that are currently being worn. You can get the current contents of the COF with the below:
```SQL
SELECT
	*
FROM
	inventoryitems invi
WHERE
	invi.parentFolderID = (SELECT folderID FROM inventoryfolders WHERE agentID = 'ee1602c0-ae2c-4867-a893-509f6c88ba0b' AND type = 46)
```
If the above SQL has more than one entry of AssetType 25 then you know that your viewer's been leaving a mess behind - this mess is what this patch cleans up when you change outfits.

The viewers seem to assume that there will only be one folder link in the COF, and that that is the folder link for the currently worn outfit.

jimtarber:
This change looks good, with one exception.  If old entries are removed automatically by the server, then when a well-behaved viewer requests the active outfit folder link be removed, the server should not report that as an error if it does not exist (currently it does report it).  So I think this PR include one further commit so that the Remove error should be conditional on not being a remove of a AssetType.FolderLink item.

jimtarber:
Ignore my comment, the "item not found" case is the one case in Remove that doesn't actually have a console error message.  This commit is good to go (although I think you wanted one more tweak).

PR #118, opened 2015-12-29
Fixed a dictionary add exception that would happen on any localID update

jimtarber:
There's a SetItem that can be used like a Replace anyway, instead of
calling Add/Remove, since calling them in either order leaves a time
when the dictionaries aren't consistent.  Single update to handle the
remove/add as a replace.

This is a merge into the release branch. Separate PR #117 for master.


jimtarber:
This is no good, the problem is in the caller. We need to do a separate remove of the old one anyway.  (Ignore PR.)

PR #120, opened 2015-12-31
Increases memory limit of scripts from 64k to 128k.  This is due to t…

zauberparacelsus:
…he possibility that some (large) scripts from Second Life may not function under Phlox, due to strings being utf-16 instead of utf-8.


ddaeschler:
Im going to close this so that we can build the phlox DLLs from our build machine instead

PR #122, opened 2015-12-31
Fixed a regression that caused bot profiles to not be found.

jimtarber:
The fixes for user profile caching missed a special exception case for
temp profiles which left them unable to be returned from a UUID search.
Special check restored in this commit.  Fixes missing bot names,
profiles, and ability to create a second bot with the same name, etc.


jimtarber:
For some reason, git is once again showing more than the single commit that is different. It is also showing the commit that was cherry-picked, although I wouldn't expect that to have any effect on the merge since master already has it.

PR #127, opened 2016-01-01
(Low priority) Deduplication and safe access for ScenePresence::PhysicsActor

kf6kjg:
More things being fixed even if not obviously borked because they annoy me :P


ddaeschler:
Looks good. Thanks!

PR #133, opened 2016-01-06
IW_OBJECT_SCRIPT_MEMORY_USED

zauberparacelsus:
This branch (which is misnamed) adds a new inworldz-specific constant for llGetObjectDetails and llGetAgentDetails, IW_OBJECT_SCRIPT_MEMORY_USED.  It functions like the OBJECT_SCRIPT_MEMORY constant, but instead of assuming each script is using the maximum amount of memory, it instead reports the actual memory usage of the scripts.

I could have simply changed the behavior of OBJECT_SCRIPT_MEMORY, but it occurred to me that scripts imported from Second Life may break if they're coded to assume SL behavior.

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
@@ -5086,10 +5086,8 @@ namespace InWorldz.Phlox.Engine
                             case ScriptBaseClass.OBJECT_SCRIPT_MEMORY:  // total mem possible (128K)
                                 total += engine.GetMaxMemory();
                                 break;
-                            case ScriptBaseClass.OBJECT_SCRIPT_MEMORY_USED:
+                            case ScriptBaseClass.IW_OBJECT_SCRIPT_MEMORY_USED:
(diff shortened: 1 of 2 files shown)


jimtarber:
One minor thing and one concern that I think I'm okay with as-is:
- we usually try to introduce new IW_ constants in the 11000 range, thinking it's less likely to be used by LL in the future. However the 10000 range is also very unlikely and we used that for the vehicle type constant extensions. For the future, try to introduce new values in the 11000-range, but no need to commit an update here.
- I'm still a bit concerned that the primary usage of this feature is to identify the users in a crowd with the highest memory usage, i.e. user shaming. I didn't really get an answer on what the goal of providing this was.  Most users are not scripters, they are just users of products, and I don't like the idea of club owners complaining about certain users or attachments that happen to make use of more of the system available. Phlox is very efficient and identifying the _relative_ highest usage provides little benefit, if any.  That said, the reason I'm okay with this one is that the existing SL-compatible option to do this returns even larger numbers.  So I don't think this makes the problem worse.  As a result, I'm okay with this commit.  I've also reviewed the code, it looks good.

jimtarber:
Looks like this one got missed in .24, so if there aren't any further objections, I think this one is ready to be merged.

PR #148, opened 2016-02-04
Fixed bot outfit changes to not interact with the bot owner's outfit if in the same region.

jimtarber:
Bot outfit items no longer share the same (real actual) user inventory item ID with the bot owner (and other bots). Instead, bot outfit items are assigned a new unique inventory item ID when they are worn, and that ID is mapped backed to the original inventory item ID.  Fixes [Mantis #3164](http://bugs.inworldz.com/mantis/view.php?id=3164)


jimtarber:
PR updated to specifically avoid all commits of bot attachments, since it was getting queued as an asset update but then being skipped farther down after a lot of needless code ran.

jimtarber:
Okay, conflicts resolved. There was just 1 conflict in BotManager.cs but GitHub shows other changes already committed to master merged back in here. Sigh.

PR #149, opened 2016-02-05
Fixed null reference exception in llGetLandOwnerAt, others.

jimtarber:
Fixed null reference exception in **llGetLandOwnerAt** when parcel coordinates are outside region. See [forum discussion](http://inworldz.com/forums/viewtopic.php?f=12&t=22214) for more.

It was a pattern so _several_ places updated. Also improved some of the error message cases to indicate the location (coordinates) in question. 

See also [Mantis #3170](http://bugs.inworldz.com/mantis/view.php?id=3170)


ddaeschler:
Looks good. Merging.

PR #154, opened 2016-02-10
Cloud assets cache: Auto-repair the empty cache case

jimtarber:
This auto-repairs the case where we should have a locally-cached Stratus
asset but we don't have the cached item because the disk was full so the
cached item is 0 bytes (or missing).

Not tested: I don't have a local CloudFiles setup. Will test on beta.

### DIFF
--- a/InWorldz/InWorldz.Data.Assets.Stratus/Cache/DiskWriteBackCache.cs
+++ b/InWorldz/InWorldz.Data.Assets.Stratus/Cache/DiskWriteBackCache.cs
@@ -170,6 +170,34 @@ namespace InWorldz.Data.Assets.Stratus.Cache
         }
 
         /// <summary>
+        /// Auto-repair the case where we don't really have a cached item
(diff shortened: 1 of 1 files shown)


InWorldz/InWorldz.Data.Assets.Stratus/Cache/DiskWriteBackCache.cs
```diff
@@ -216,6 +252,15 @@ public StratusAsset GetAsset(Guid assetId)
                     return null;
                 }
 
+                // Auto-repair the case where we don't really have a cached item
+                FileInfo fileInfo = new System.IO.FileInfo(GetAssetFileName(assetId));
```
ddaeschler: Is this fileInfo ever used?

jimtarber:
I've fixed the unnecessary declaration which was left behind after refactoring some common code into RepairEmpty().

PR #155, opened 2016-02-11
Implemented iwValidateURL function, which allows a scripter to verify…

zauberparacelsus:
… if a URL is formatted correctly according to the HTTP, HTTPS, and WWW schemes.


ddaeschler:
This seems to correct something like www.google.com to http://www.google.com, but not google.com or something like subdomain.google.com to also add http. Maybe the implementation should assume anything.anything is a valid domain and add the http or https? Or possibly the implementation should assume nothing about the protocol.

The point I'm making is that www is a commonly used subdomain/CNAME, but there are a lot of other valid domain URLs that won't begin with www that could also use the substitution, so I'm not sure about the way www is handled.

zauberparacelsus:
Aye, some of the issues were discussed yesterday.  When I'm next on my windows drive, I'll be replacing this with a fairly comprehensive regex that Cinder wrote for validating URLs.

ddaeschler:
Awesome. Looking forward to it.

jimtarber:
I think the simple URL resolution is to test for a scheme(protocol) name (e.g. "http"), followed by specifically "://" and then something consecutive with a dot in it. Can get clever with requiring the period somewhere other than at the beginning or end.  But all of that is based on us reinventing what has already been implemented.

Check the documentation for _System.Uri_ which probably has everything we need. For example, it has a _Uri.SchemeDelimiter_ definition, methods to split URIs into components (where we could check for our own patterns to recognize URLs. For example, if System.Uri only recognized _specific _schemes, we could still use it to split the URL into scheme, authority(domain), and path+parameters, and then apply our own checks for the components. It also handles escaping in both directions. More specifically, it has _CheckSchemeName_, _CheckHostName_, _IsWellFormedUriString_, etc.

But it may be as simple as calling the _Uri(string)_ constructor and catching exceptions.  See [the Uri Constructor documentation](https://msdn.microsoft.com/en-us/library/z6c2z492%28v=vs.110%2

zauberparacelsus:
This pull request was updated, and includes an additional set of misc fixes:

* I removed the www test from iwValidateURL.  Also, Cinder's regex didn't work for some reason.
* Improved some of the argument handling in iwStringCodec to make some of the arguments case-insensitive.
* Made CHANGED_INVENTORY trigger in the root prim when a child prim's inventory is changed (matching SL behavior)
* Stop CHANGED_INVENTORY from triggering when no-copy items were removed (matching SL behavior)
* Stop CHANGED_TEXTURE from being triggered by scripted texture, glow, color, shine, texgen, etc. changes (matching SL behavior)

PR #156, opened 2016-02-12
Object part flags

jimtarber:
The fixes that were in #152 were undone by this polluting of the Scripted flag from the root prim. This PR just preserves the Scripted flag so as to provide the fix from #152, while *not* applying it to all other flags, as they may have different semantics.


jimtarber:
Actually going to do a PR from the .21 release branch instead (same thing)

PR #171, opened 2016-03-02
Release 0.9.22

jimtarber:
If #169 is merged, this one should have no effect, but then shows as merged.


jimtarber:
This now includes a NRE fix that is not already merged in master, so this PR merge would have an important result.

jimtarber:
Now includes the fix for llEmail-triggered exception that halted the script engine on I'z Straits 8.

PR #175, opened 2016-03-07
Release 0.9.22 (now includes more friend perms fixes)

jimtarber:
Fixes to friend perms caching and notifications, fixes to which is checking which on friend perms lookups, fixes to friend Edit and friend online visibility notification checks.  Includes the safer required fixes up til March 6 but **not** the profile lookup-limiting changes from March 7 (today).


jimtarber:
This now includes the profile fetching limit fix.  Should resolve the DoS-like request flood seen on the Sandbox regions caused by many profile requests occurring during the window while a cache miss was fetched.  GitHub doesn't show the diff well due to whitespace (indenting) changes. Can view the diff without them at https://i.gyazo.com/163e96cfc9ab74b85bb15ff33a9cf826.png and https://i.gyazo.com/9bb7023fe7e5263cfad5add7a6f8b2f0.png

PR #181, opened 2016-03-12
Ll remote load script pin error tweak

zauberparacelsus:
This Pull Request contains two changes two llRemoteLoadScriptPin:

1) If llRemoteLoadScriptPin's PIN argument is zero, then instead of trying to use that PIN as though it were valid, it will instead throw an error saying _"llRemoteLoadScriptPin: PIN cannot be zero"_ and then do an early return (with a 3000ms sleep, as it would normally do with or without a valid PIN)

2) If the target prim's PIN mismatches the one passed to llRemoteLoadScriptPin, then instead of simply giving an error stating _"PIN mismatch"_, it will alternatively say _"PIN not set"_ if the target prim's PIN is zero.  

Combined, these changes should give scripters a bit more useful information when debugging scripts that are calling llRemoteLoadScriptPin and llSetRemoteScriptAccessPin

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
@@ -8702,6 +8702,11 @@ namespace InWorldz.Phlox.Engine
 
         public void llRemoteLoadScriptPin(string target, string name, int pin, int running, int start_param)
         {
+			if (pin == 0) {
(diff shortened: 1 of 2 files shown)


jimtarber:
This looks great. Great idea for helping scripters identify why their script isn't working without just telling them "something" is wrong.

ddaeschler:
Merged, thanks!

Issue #145, opened 2016-01-20
Get rid of insane overloads on SimChat


ddaeschler:
Inside Scene.PacketHandlers.cs, there are an insane number of confusing overloads on the SimChat function. These have already caused a regression and are prone to cause more coding errors especially if the method signatures need to change again in the future.

I propose that only the first SimChat method that takes all of the parameters be left in place and any parameters that now completed as part of the overloads be left as optional parameters with sane defaults.

ddaeschler:
Full path to the file is OpenSim\Region\Framework\Scenes

ddaeschler:
No takers for this? Definitely a more simple way to get your feet wet in the code

MatthewBeardmore:
Pull request #191

jimtarber:
These changes were included as part of the 0.9.23 release. ;)

PR #191, opened 2016-03-20
Get rid of insane overloads on SimChat #145

MatthewBeardmore:
Clean up SimChat method and make much simpler methods for it that have proper documentation as to what each argument does. Get rid of "Sender" property in OSChatMessage as OSChatMessage does not need to have an IClientAPI reference at all times.


ddaeschler:
Much better. I don't see any obvious issues and this is an improvement.

jimtarber:
These changes were included as part of the 0.9.23 release. ;)

PR #193, opened 2016-03-23
Changed inventory fixes

zauberparacelsus:
Although misnamed, this contains my fixes to how CHANGED_INVENTORY and CHANGED_TEXTURE are triggered.

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
@@ -2140,7 +2140,7 @@ namespace InWorldz.Phlox.Engine
                     textureChanged = true;
                 }
                 if (textureChanged)
-                    part.UpdateTexture(tex);
+                    part.UpdateTexture(tex, false);
                 return;
             }
         }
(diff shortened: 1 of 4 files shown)


jimtarber:
I'm concerned about some of the changes in this one. They change the semantics of existing functions in a way that isn't backwards compatible. And worse, the changes are unintuitive (may represent bugs in SL, or at the very least, unexpected semantic quirks). I'm not sure we want to emulate most of these changes.

jimtarber:
This now includes a further tweak to the semantics and return codes after good feedback from Wolf.

appurist:
Is there anything that should be salvaged from this PR?

One concern I had was that you cannot actually see the places affected by the changes in the diff, because UpdateTexture was changed to accept a boolean but  it has a default parameter of `false` which means that only the places _not_ affected are documented with the `,true` change to the caller. But that is only a concern about the review process.

My main technical concern here is that there would be texture changes that formerly triggered a `CHANGED_TEXTURE` message that no longer do.  

I'm also concerned by the change to notify the root prim of `CHANGED_INVENTORY` when the inventory in that prim has not actually had a change.

The change to eliminate `CHANGED_INVENTORY` notifications on persistence restores may be a good one to apply if it can be confirmed that only the persistence restores are affected.

I'd like to figure out if there's anything to proceed with on this one and if so, apply it, otherwise close  the PR (defer indefinitely, as I have done for some of mine), rather than leave it hanging as an open PR.

PR #194, opened 2016-03-25
Make sure that bots do not ever get information put into SceneView as…

MatthewBeardmore:
… they can't use it. Also make sure to fully clear out SceneView in case it isn't disposed properly by the GC later.


ddaeschler:
Looks good. Thanks!

jimtarber:
There may be a problem with this commit; when bots first appear in a region, it seems no updates are sent, the bot does not rez for the owner standing only 2m away. (If I roll the mouse wheel to zoom out/in, they appear.)

jimtarber:
I have confirmed the problem and fixed it (plus some other additional IsBot checks) in a follow-up PR #197

PR #198, opened 2016-03-29
iwRemoteLoadScriptPin

zauberparacelsus:
This is a variant of llRemoteLoadScriptPin, which returns an integer status code instead of shouting an error message when the PIN is incorrect or the target has no PIN set.  This allows scripts to detect a PIN issue and respond to it, and also avoids irritating people with script errors.


jimtarber:
This looks good. Ready to merge.

PR #203, opened 2016-04-05
Fixes to outgoing HTTP request headers from LSL scripts

jimtarber:
- Fixed the format of the default User-Agent header for LSL scripts calling llHTTPRequest, which was not accepted at some servers, notably Shoutcast servers. They would reject calls (with a 403) from InWorldz scripts that would work unchanged in SL. Fixes Mantis 3194.
- Fixed the llHTTPRequest to allow User-Agent overrides, i.e. HTTP_CUSTOM_HEADER, "User-Agent" is now supported. (Actually it now allows any non- X-secondlife header to be overridden, although there's only one at the point this is called, unless the caller specifies the same header more than once, now the last one wins.)
- Added support for more LSL HTTP request headers: "accept", "content-length", "content-type",
"expect", "host", "date", "if-modified-since", "range", "transfer-encoding" (auto-enabling chunked sends), and "connection" (including recognizing keep-alive).


jimtarber:
Don't merge yet, going to do some SL-compatibility updates Mike suggested.

PR #218, opened 2016-05-14
Release 0.9.24

jimtarber:
This is the 0.9.24 branch merged back into master.

I believe this will be the contents of #213 and #215 along with the one addition commit to _revert_ the restriction against moving no-copy items out of no-mod object Contents.


jimtarber:
If this PR is applied, we can ignore #213 and #215, this one includes those.

PR #220, opened 2016-05-22
Fixes to ensure region backups are made, fixes "(RSF) Rookery" problem.

jimtarber:
- If the specified region cannot be found, instead of throwing an exception, report the error but default to the actual current region/scene so that a backup is done.
- Fixed case-sensitive region name compares by replacing most of FindSceneByName with the usual TryGetScene function.
- Added *.oarstatus to gitignores

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -250,3 +250,4 @@ bin4
 *.oar
 bin/blacklist.old
 bin/blacklist.txt
+*.oarstatus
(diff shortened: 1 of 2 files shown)


jimtarber:
After discussing this with @ddaeschler earlier, I've learned that the current behavior is at least partly intentional; this commit is on hold until we sort out the desired behavior.

PR #235, opened 2016-06-24
Automatic map tile export capability

kf6kjg:
This adds the ability to, if configured to do so, export the map tile to a specific file location when the region detects that the map has been changed.  That file location could be on a network share to the map server, or another tool could be used to copy the files to the map server, where they can then be digested by the supertile generation system (cough, Anaximander, cough) to create the full map set.

The code has been written to accept existing Halcyon.ini configurations.  Such legacy configurations will simply result in this new system being not activated.  To be  clear: updating a region to a version with this patch does not require changing the existing Halcyon.ini.  Only if you want to activate map exporting will you need to change the INI.

Please note that there has been a purposeful semantic change: only non-temp non-physical objects are now part of the map rendering.  The idea being that objects that are flagged physical and/or temporary are transient and therefore shouldn't be part of a "map" anyway.


kf6kjg:
Included the optimization that makes the process significantly faster.

jimtarber:
This PR looks good to me, pending good answers to the following: 
- Is there a plan to have MinimumTaintedMapTileWaitTime in the ini file? 
- What is the typical MinimumTaintedMapTileWaitTime (minimumMapPushTime).  It's important that these never collide as it could cause a backlog in the FireAndForget threads.

I'm thinking it might be good to ensure a _minimum_ minimum that should be capable of always exceeding the longest time for a ExportMapTileToDisk call.  That's probably only seconds but the minimum time is specified in seconds so someone putting a 1 in there could result in bad juju. ;)  Maybe it should ensure a minimum of 10 minutes between these calls or something like that.

kf6kjg:
Already there, but commented out as it sets the default:
;MinimumTaintedMapTileWaitTime = 3600

MaximumTaintedMapTileWaitTime is set with a "sane" value as it has a
default of 0 = "every time the map is tainted".

I did a worst-case test on my Linux box, and the regions handled it -
however it really put a strain on the machine, and the region only had a
few scripted object and only one avvie.  As this is set by the
administrator of the grid, I didn't want to initially limit the minimum
other than make sure it was not negative.  My thoughts were: if the
sysadmin want to kill his server, just make sure he's been warned. :D

jimtarber:
I don't follow, because these comments seem self-contradictory. What is the default, is it 3600 or 0? If it's 3600, I think this PR is good to go.

> Already there, but commented out as it sets the default: ;MinimumTaintedMapTileWaitTime = 3600

> MaximumTaintedMapTileWaitTime is set with a "sane" value as it has a default of 0 = "every time the map is tainted".

Is this not new code (the default in previous releases was to never do this)? So the administrator doesn't know this is coming and the default should be to either not do this at all or do it infrequently enough so as to not add load to the system.

I'm not actually all too concerned with load, although that is a concern. What I'm worried about is every change causing a new thread to be added and blocked on the Monitor lock and a backlog in FireAndForget threads created that just builds and builds.  If every prim change, terrain edit, etc,  immediately fires up a new thread (by default), we have a serious problem here. If it only checks once every 3600 seconds, it's not a problem at all.  

I think a default of 3600 seconds is good.  If that is what it is, then this PR is in good shape. If it's 0, that needs to change.

Separately, I thiunk we should probably also have a hard minimum of something like 60 seconds (or even longer if map generation could take longer than that).

kf6kjg:
As noted in the INI file diff comments, unless RegionMapTileExportFilename is set - which it is not by default - none of the core of the new code is operational especially none of the disk writes.   Thus no surprises for existing regions or grids.

I think there's some confusion here, and it may be due to the simplistic naming of the entries.  While MaximumTaintedMapTileWaitTime and MinimumTaintedMapTileWaitTime are simple handles that a sysadmin can comprehend easily, they are not treated directly as max and min in the code.  From a developer's POV the "Min" means "the amount of time needed to pass before allowing a disk write of the map tile" while "Max" means "the interval in seconds that a timer should check to see if the map has been tainted and that Min time has passed."

Thus if Max < Min, it simply means a timer will be firing faster than it really needs to be, but most of the time it'll be just hitting a falsy if statement and dropping out immediately.  If Max == 0, then the timer never fires.  If Min == 0, then every map taint will initiate a F&F for the purpose of doing the disk write.  However if there is already a F&F disk write in progress, then the new F&F fails the TryEnter call and aborts out rapidly.  These files are small: 256x256px JPEG images.  They write fast to local disks, and fairly fast to network disks.  A network disk could take a long time to fail/succeed, in which time multiple F&F's could have hit the lock and aborted - not a huge waste as hitting the lock and aborting is VERY fast.

Now, the value given to "Min" in the sample INI is not the "default" - where default means what the code will assign if you do not give a value.  It is instead a casual recommendation, and a reasonable starting point for those who want to try and experiment with activating this block of code.

Map generation takes milliseconds on my test region.  I do want to do an insanity test though: what happens if the space from ground level to ground level + 256 if filled to region capacity with prims that are 1x1x1 meter or larger?  How long does it take?  I'll see if I can run a test on that! :D

kf6kjg:
Constants - they vary from run to run:
* Generating terrain image takes 40 ms.
* Encoding image as JPEG2000 takes 40 ms.
* Attempting to save to asset server, and failing, takes 100 ms.
* Disk write of map tile to my platter disk takes 40 ms.

There are two speed tests at two object counts: I have the code as-is in this patch, and the code after an as-yet unreleased optimization to the drawing code; I also have two object count extremes.

Code as-is, 11 objects in drawing area: 1 ms.
Code optimized, 11 objects in drawing area: 1 ms.
Code as-is, TODO objects in drawing area: TODO ms.
Code optimized, TODO objects in drawing area: TODO ms.

(I will be editing this comment with the high prim count results as soon as I can - ran out of time tonight)

jimtarber:
> Thus if Max < Min, it simply means a timer will be firing faster than it really needs to be, but most of the time it'll be just hitting a falsy if statement and dropping out immediately. If Max == 0, then the timer never fires. 

I wasn't concerned about anything to do with the Max value, only where Min < time to process. However...

> If Min == 0, then every map taint will initiate a F&F for the purpose of doing the disk write. However if there is already a F&F disk write in progress, then the new F&F fails the TryEnter call and aborts out rapidly.

When I first reviewed this, I missed the fact that it was a **_Try_**Enter which is why the minimum time mattered greatly.  Now that you've pointed out that it is a _conditional_ Enter, that significantly lowers my concerns.

That said, I am still concerned by the low minimum (0 in this case).  We really don't need to ever support such low numbers, and I don't really see any value at all in supporting a minimum time of 0. Almost any larger value avoids any hint of a problem.  

Let me spell out my concern, even with the TryEnter: A value of 0 in a fast-changing region effectively means tight loop, and if changes occur at a rate that is faster than the time to process, a full CPU could still be used.  If changes happened only every 100ms, and processing took 50ms, we could see 50% CPU usage.  (And 10 changes per second is _very_ low compared to what we've seen.)  In a region where there are moving objects and other changes, this could cause a lot of thread use just firing up the FireAndForget handlers that did nothing much. I really think the minimum should be limited to at least a minute, preferably something more like 10 minutes. And it certainly shouldn't be the default to 0, as we _know_ that will be shorter than the processing time.

jimtarber:
I didn't really understand the answer to this:
> What is the default, is it 3600 or 0? If it's 3600, I think this PR is good to go.
So I checked the code again and it looks like the default is 3600, not 0:
```private TimeSpan minimumMapPushTime = new TimeSpan(1, 0, 0);```
and as long as the Max default of 0 doesn't cause that to change, and this code:
```
int pushTimeSeconds = Math.Max(0, worldmapConfig.GetInt("MinimumTaintedMapTileWaitTime", (int) minimumMapPushTime.TotalSeconds));
minimumMapPushTime = new TimeSpan(0, 0, pushTimeSeconds);
```
So if that ends up with 3600 in pushTimeSeconds by default, then I think the defaults are all good.

However, since it is configurable, I'd like to see a minimum non-zero value enforced for that pushTimeSeconds (which is basically a minimum for MinimumTaintedMapTileWaitTime) so that administrators can't shoot themselves in the foot.  Unless they install all the .PDB files and know how to pull stack traces on all threads and understand what they mean, this could result in a nasty performance problem that is difficult to diagnose.  Avoiding that is as easy as adding a check for pushTimeSeconds' value after it fetches MinimumTaintedMapTileWaitTime. How does a hard 1 minute limit sound for that?  (Do we really need to push updated maps more frequently than once per minute?)

kf6kjg:
60 seconds as a hard limit for Min sounds good to me.

jimtarber:
All looks good, thumbs up on the PR. Ready for merge.

kf6kjg:
As of Anaximander 2 this becomes a moot point for me: Anax2 generates the map tiles itself and does a better job at it.  Maybe it'd help someone, but for now I'm just closing the PR as my purpose for this is gone.  At some point I'll get Anax2 server mode working and submit a PR to connect regions to that for nearly-live map updates, but that won't be this code.

PR #236, opened 2016-06-25
Fix an issue with a null value on a JSON object when calling llJson2List.

mdickson:
Added Test support for the Phlox scripting Engine including a Mock for the Engine. Added JsonTests for lsl script engine tests around JSON handling.  Added an initial failing test cases and some positive cases for llJson2List handling an json object with a null value for an element.  Added code to get the failing test passing.


jimtarber:
Looks great. The ```nested``` check is critical and included in the commit, so these changes look great and open the door to future test cases with the unit testing wrapper also included.  PR ready for merge.

PR #239, opened 2016-07-01
Exit code differentiation between "restart me" and "don't restart me"

kf6kjg:
Now crashes are exit code 1, restarts from any source are exit code 100, and ONLY an explicit shutdown is exit code 0.

### DIFF
--- a/OpenSim/Framework/Servers/BaseOpenSimServer.cs
+++ b/OpenSim/Framework/Servers/BaseOpenSimServer.cs
@@ -311,7 +311,8 @@ namespace OpenSim.Framework.Servers
         /// </summary>      
         public virtual void Shutdown()
         {
-            Shutdown(0);
+            // Use a status of 64 (0b01000000) to indicate that this "error" is an explicit shutdown and not a real error.
+            Shutdown(64);
         }
 
         /// <summary>
--- a/OpenSim/Framework/Servers/BaseOpenSimServer.cs
+++ b/OpenSim/Framework/Servers/BaseOpenSimServer.cs
@@ -311,12 +311,21 @@ namespace OpenSim.Framework.Servers
         /// </summary>      
         public virtual void Shutdown()
         {
+            Shutdown(0);
+        }
+
+        /// <summary>
+        /// Shutdown the server with the specified exit code.
+        /// </summary>
+        /// <param name="exitCode">The exit code to be returned once shutdown has completed.</param>
+        public void Shutdown(int exitCode)
+        {
             ShutdownSpecific();
-            
+
             m_log.Info("[SHUTDOWN]: Shutdown processing on main thread complete.  Exiting...");
             RemovePIDFile();
-            
-            Environment.Exit(0);
+
+            Environment.Exit(exitCode);
         }
 
         private void HandleQuit(string module, string[] args)
--- a/InWorldz/Halcyon/Application.cs
+++ b/InWorldz/Halcyon/Application.cs
@@ -65,6 +65,9 @@ namespace OpenSim
         //could move our main function into OpenSimMain and kill this class
         public static void Main(string[] args)
         {
+            // Under any circumstance other than an explicit exit the exit code should be 1.
+            Environment.ExitCode = 1;
+
             // First line, hook the appdomain to the crash reporter
             AppDomain.CurrentDomain.UnhandledException +=
                 new UnhandledExceptionEventHandler(CurrentDomain_UnhandledException);
--- a/OpenSim/Base/OpenSim.cs
+++ b/OpenSim/Base/OpenSim.cs
@@ -778,8 +778,8 @@ namespace OpenSim
 
                 case "restart":
                     // Let's issue a full server shutdown and let the management tools take care of the restart.
-                    // Use a status of 100 to indicate that this "error" is an explicit restart not a real error.
-                    Shutdown(100);
+                    // Use a status of 63 (0b00111111) to indicate that this "error" is an explicit restart request and not a real error.
+                    Shutdown(63);
                     break;
 
                 case "predecode-j2k":
--- a/OpenSim/Base/OpenSim.cs
+++ b/OpenSim/Base/OpenSim.cs
@@ -777,7 +777,9 @@ namespace OpenSim
                     break;
 
                 case "restart":
-                    m_sceneManager.RestartCurrentScene();
+                    // Let's issue a full server shutdown and let the management tools take care of the restart.
+                    // Use a status of 100 to indicate that this "error" is an explicit restart not a real error.
+                    Shutdown(100);
                     break;
 
                 case "predecode-j2k":
--- a/OpenSim/Grid/GridServer/GridServerBase.cs
+++ b/OpenSim/Grid/GridServer/GridServerBase.cs
@@ -115,7 +115,7 @@ namespace OpenSim.Grid.GridServer
                     System.Threading.Thread.Sleep(delay * 1000);
 
                 // Do this on a new thread so the actual shutdown call returns successfully.
-                Task.Factory.StartNew(() => Shutdown());
+                Task.Factory.StartNew(Shutdown);
             }
             catch (Exception e)
             {
--- a/OpenSim/Grid/GridServer/Program.cs
+++ b/OpenSim/Grid/GridServer/Program.cs
@@ -25,6 +25,7 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+using System;
 using System.Net;
 using log4net.Config;
 
@@ -34,6 +35,9 @@ namespace OpenSim.Grid.GridServer
     {
         public static void Main(string[] args)
         {
+            // Under any circumstance other than an explicit exit the exit code should be 1.
+            Environment.ExitCode = 1;
+
             ServicePointManager.DefaultConnectionLimit = 12;
 
             XmlConfigurator.Configure();
--- a/OpenSim/Grid/MessagingServer/Main.cs
+++ b/OpenSim/Grid/MessagingServer/Main.cs
@@ -159,7 +159,7 @@ namespace OpenSim.Grid.MessagingServer
                     System.Threading.Thread.Sleep(delay * 1000);
 
                 // Do this on a new thread so the actual shutdown call returns successfully.
-                Task.Factory.StartNew(() => Shutdown());
+                Task.Factory.StartNew(Shutdown);
             }
             catch (Exception e)
             {
--- a/OpenSim/Grid/MessagingServer/Main.cs
+++ b/OpenSim/Grid/MessagingServer/Main.cs
@@ -65,6 +65,9 @@ namespace OpenSim.Grid.MessagingServer
 
         public static void Main(string[] args)
         {
+            // Under any circumstance other than an explicit exit the exit code should be 1.
+            Environment.ExitCode = 1;
+
             ServicePointManager.DefaultConnectionLimit = 12;
 
             XmlConfigurator.Configure();
--- a/OpenSim/Grid/UserServer/Main.cs
+++ b/OpenSim/Grid/UserServer/Main.cs
@@ -80,6 +80,9 @@ namespace OpenSim.Grid.UserServer
 
         public static void Main(string[] args)
         {
+            // Under any circumstance other than an explicit exit the exit code should be 1.
+            Environment.ExitCode = 1;
+
             ServicePointManager.DefaultConnectionLimit = 12;
 
             PIDFileManager pidFile = new PIDFileManager();
--- a/OpenSim/Region/Framework/Scenes/Scene.cs
+++ b/OpenSim/Region/Framework/Scenes/Scene.cs
@@ -998,7 +998,7 @@ namespace OpenSim.Region.Framework.Scenes
         /// Given float seconds, this will restart the region.
         /// </summary>
         /// <param name="seconds">float indicating duration before restart.</param>
-        public virtual void Restart(int seconds)
+        public override void Restart(int seconds)
         {
             // RestartNow() does immediate restarting.
             if (seconds == -1)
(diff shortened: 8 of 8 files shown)


OpenSim/Framework/Servers/BaseOpenSimServer.cs
```diff
@@ -311,12 +311,21 @@ public virtual void Startup()
         /// </summary>      
         public virtual void Shutdown()
         {
+            Shutdown(0);
```
kf6kjg: There is debate as to whether or not to exit with status 0 on a shutdown.  The arguments are:
* 0, and possibly any value from 0 to 50, are very common exit codes.  A library could call exit with that value at any time and defeat the whole purpose of being able to disambiguate a shutdown-and-don't-restart from errors or restart operations that should automatically come back up.  To choose 0 would be to leave ambiguity about whether or not to restart.
* 0 is commonly used to mean "exited successfully" and a shutdown without intent to restart is a successful and proper termination of Halcyon.  To choose some other value and leave 0 for errors is contradictory to standard practice and common expectation.

Either choice that is made will lock in the meaning and cannot be changed in the future as tools will come to rely on the value.

kf6kjg:
As noted in the diff:
There is debate as to whether or not to exit with status 0 on a shutdown. The arguments are:

* 0, and possibly any value from 0 to 50, are very common exit codes. A library could call exit with that value at any time and defeat the whole purpose of being able to disambiguate a shutdown-and-don't-restart from errors or restart operations that should automatically come back up. To choose 0 would be to leave ambiguity about whether or not to restart.
* 0 is commonly used to mean "exited successfully" and a shutdown without intent to restart is a successful and proper termination of Halcyon. To choose some other value and leave 0 for errors is contradictory to standard practice and common expectation.

Either choice that is made will lock in the meaning and cannot be changed in the future as tools will come to rely on the value.

kf6kjg:
There is also a third argument that no external process should be relying on Halcyon's exit codes for decision making.  Those tools should be used by the sysadmin to initiate shutdown if wanted and can therefore already know not to restart the region once it is down.

jimtarber:
From a purely "proper coding" perspective, we shouldn't be changing the server code at all for this. However I recognize it is the most pragmatic choice among the possible approaches in terms of cost/benefit of effort, and think there's a compromise here. I think it's more practical on some grids to be able to use the shutdown console command, and for that reason, I think exit codes are a reasonable approach, even if it's not the most proper.

That said, I suggested changing the shutdown code to practically _any_ other value _other than_ 0, because 0 is the one value in the whole 32-bit range that is _most_ likely to be used on exit by some other unknown code, thus it is the worst possible choice.  The fact that 0 is a standard/default exit code in fact strengthens this point; it is the most likely value to be problematic, and the result of it not being unique to our shutdown needs results in regions remaining down when they should have been restarted. This is the one and only exceptional case where grids might want to stray from the _normal termination_ behavior (which is to restart the region), and using 0 here is asking for trouble.  The cost/benefit of choosing 0 when it can be any value is an almost infinite ratio.

We don't need to take risks here. I cannot support this PR with shutdown==0.  We need a unique exit code value for the exceptional case, and 0 is the _least_ likely to be unique.

I suggest the simple tweak of changing shutdown to _something_ else (such as 99).  The restart exit code could remain as 100, or since the normal behavior is to try to keep regions up and running and restarting a region on an exit is the norm, restart _could_ return a 0 exit code. However, I think it's best to leave restart at a specific value however, and the 100 value used in this PR is fine.  But by using specific values for both, we could actually log _the reason_ for termination, even if we treat them all the same (e.g. even in the InWorldz case where we will almost always want to do a restart).

OpenSim/Framework/Servers/BaseOpenSimServer.cs
mdickson: IMO a clean exit (shutdown OR restart) should return 0.  Anything else is a non-zero return.   Where anything else is generally reserved for a crash or unusual termination (process killed).  Returning "state" in an exit code now requires a caller to check it.  And IMO that's bad practice.

jimtarber: Technically both restart and shutdown are clean exits so yes, I agree the proper return value would be a 0.  However, if a 0 was returned as in the past, it be ignored by all callers, and not allow the admin to use the ```shutdown``` console command without aggravation. The whole purpose of this PR is to allow the ```shutdown``` console command to be used rather than an external system.  In both cases, the region is notified to shutdown by a human operator.  The region is not making any decisions, it is the recipient of the command from the human operator.

This PR attempts to allow any callers to _optionally_ recognize the _abnormal_ case of a manual intervention by the human, requesting an exit that should **not** be turned into a restart like it normally would.  That exceptional case is the direct result of use of the shutdown command on the console.  All other terminations, graceful or crashes, would result in a restart.

Because Halcyon _supports_ the ```shutdown``` console command (at all), I don't have a problem with differentiating that on the exit, if it allows the _operator's_ intentions to be recognized. If we don't, then the alternative, really, is to remove the shutdown command, since we're effectively saying don't use it for a shutdown where you _don't_ want a restart. I don't think that is practical and I don' think it's a philosophical crime to _allow_ the operator to manually intervene. However, I do think it would be a philosophical crime to _require_ the operator to intervene.

And that is why I don't like the exceptional case of a shutdown (no restart) returning a standard exit code, especially the most common 0 case. It is an exceptional request of manual intervention by an operator to bring a region down and intend for it to stay down, rather than to allow any controlling software to restart it automatically (including a batch script in a loop).  It is an exceptional case and we don't need to pick the _most normal_ of exit codes.

However, the last two sentences above aren't correct; no exit code requires a caller to check it. For example, neither Zookeeper nor Maestro would check it.  It only _allows_ the grid owner to differentiate if convenient to their process, as is the case on GW grid, which is what triggered this PR.

kf6kjg:
I apologize for my bullheadedness and pride.  As you can see, I've implemented such a change to the patch.

jimtarber:
I'm satisfied with this PR as a whole now; having additional explicit exit codes for specific operations allows the invoker to either differentiate them _or_ treat them the same, where differentiating may involve behavior differences (like at GW) or not (like at IW), or may only result in different logging. (For example, we might want to indicate in ZK or Maestro that the process exited with a specific exit code, even if we will always be restarting it.)

From my perspective, this PR can be merged now.

appurist:
@kf6kjg is asking about this, and I think it's good to go. It doesn't cost anything in the managed cases like InWorldz' ZK/Maestro use, but provides a positive definitive mechanism for the simpler grid setups.

mdickson:
Maestro if I recall only checks for a non-zero exit code assuming that means an error occurred. There is only support via the admin api for a shutdown.  If a process is restarted the assumption is the management engine is doing it.  So if we implement some sort of special meaning for exit codes its not going to be respected or utilized at least from a system like Maestro that assumes its managing the instances.  IMO if we want to implement this then we should teach the server instance to do a soft restart.  Shutdown should mean exactly that.   And on a restart the process shouldn't exit but restart in place.  If that's not possible then we shouldn't be adding commands for it since there's nothing actionable in the process that can implement the behavior. It depends on an external system that may or may not be there to work...

kf6kjg:
Soft restart, aka reload, would be nice but is orthogonal.  Currently exit codes are somewhat chaotic; while non-zero currently means error crash, zero is ambiguous: all of crash, restart, and shutdown emit that code. 

Right now I'm working on a simplified management system based on making the code run as a daemon under Linux. The management tools there will send a sigint to tell it to shut down, which thanks to another PR now works, and will then not restart it. However if a shutdown is sent from one of the control APIs, there's no way except through exit codes for the daemon management software to know to actually stop and not restart it. 

Remember this or isn't about restart. It's about differentiating shutdown from crash/restart in an environment where the process that sends the shutdown command can't talk to the daemon manager system at all.

mdickson:
If a shutdown is being sent from a control api then it already knows what its trying to do.  Auto-restarting things is the job of the management layer. If you're using something like running as a windows service (or a Linux service definition) and you don't want the service to autorestart then stop the service.   The region service itself shouldn't be involved in decisions like that.

appurist:
There are competing use cases involved here. There is the managed ZooKeeper/Maestro-like cases where region exit was initiated by the controller, and then there are the unmanaged cases such as the adminstrator typing "shutdown" at the console prompt.

I've been trying to understand Mike's objection here and I think it may be related to the fact that "shutdown" currently generally means "restart" in most cases to the caller, at least at InWorldz in both ZooKeeper and Maestro.  When managed, even typing "shutdown" on the console means restart.  So there is a philosophical argument for not returning an code that implies an intentional shutdown.

However, the fact that it is not universal means it is up to the caller to interpret the meaning.  By "not universal", I mean cases such as automated rollouts do _not_ just restart the region when it terminates.  This is why a "soft restart" wouldn't be appropriate here.  Also, because many user-initiated restarts are being requested as a workaround or test to diagnose some specific problem in a region, it is important for a restart to actually exit the process and begin with a fresh process.

I think the conflicting use cases of managed (ZK/Maestro) and the lighter-managed cases (e.g. where the controller is a shell command with a loop) require a compromise between the idealistic nature of a fully-managed Maestro environment and other use cases.  Part of the problem here is that ZooKeeper does a restart by sending an actual `shutdown` command to the region server console, so that adds to the interpretation of `shutdown` as a restart request.

Perhaps we should be adding a second command, a `restart` command, which has the same implementation in Halcyon, which is to simply exit the process gracefully, and update ZooKeeper to use it rather than `shutdown`.  That would provide a distinction between the cases, and might alleviate any concern over returning a result code that specifies a shutdown meaning when the common usage is to restart.  That could also possibly evolve into a soft restart for non-upgrade cases from ZooKeeper. Except that I don't think we'd ever use a soft restart for anything in practice. Neither the managed nor unmanaged use cases encourage it; if we're going to be philosophical or academic about this, the case for soft restart is the same as "don't bother issuing any commands". There should never be a need for a soft restart.

So let's be pragmatic about this and recognize that there is actually a `shutdown` console command and simpler grid setups use this all the time.  How does indicating this to the caller actually present a problem? Any termination of the process that was not expected by any management software will currently trigger a restart. Everything continues as it currently is. However, differentiating the shutdown request provides any possible caller with additional information about why it terminated. This information can be used, or as in the case of ZooKeeper/Maestro, it can be ignored and a restart initiated.

Ideally, we could eventually (or immediately) copy/paste the `shutdown` implementation and tie it to a new `restart` command and have ZooKeeper/Maestro send that instead of `shutdown`.  Or send a `shutdown` in the case of region software upgrades or operating system upgrades.  And then have each of those commands differ only in the value returned.  But that actually wouldn't make much of a difference to ZK/Maestro. It knows if it requested a shutdown or restart or upgrade, so process termination is not unexpected.  But in the case of a shell script or other wrapper that exists solely to keep the process running, it would be good to indicate if the process exit was intentional or not.  So I'm leaning towards splitting `shutdown` into two commands here, with `restart` being the same (process exit) but with a different other return value.  

The idea here is to provide a way to signal to the caller the reason for a process exit when one occurs because it may be _unexpected_. I think part of the argument against this is a claim that it _should never be_ unexpected, however that will only be true if we remove all forms of restart/shutdown other than those invoked by the management tool.  We are not going to remove the `shutdown` command, so that is not a realistic viewpoint. Regions restarts and shutdowns can be requested via means _other than_ the controlling software. However, it might be good to differentiate the two cases even further. So I think we need two commands here.

appurist:
Mike, simple question: are you suggesting that we remove the `shutdown` console command?

I would be in favor of that once the region process was converted to a Windows service and Linux daemon/service process, which ideally it should be.  But until we actually remove the `shutdown` command, it exists.  It's currently being abused by the controlling software and viewer API implementation in the sense that it is use for restarts too, but adding a second command would address that inconsistency.  Then (I think) the controlling software would only see an unexpected _shutdown_ if the user typed it on the console.

appurist:
Additional note here for the unexpected shutdown case:

Even in the case of ZooKeeper/Maestro, it would be great if the controller software recognized an _external_ request for a single-region shutdown (something ZooKeeper does not support) and did not auto-restart the process.  

This is something that causes problems for InWorldz all the time. We must manually stop ZooKeeper in order to perform maintenance on a region.  Because ZK will attempt to auto-restart the region, because it cannot tell the difference between a shutdown and a restart, because shutdown is currently overloaded to _usually_ mean restart.  However there are manual operator shutdowns too (for example, if a region is being moved or renamed or something of that nature).  We plan to be able to handle many of those cases from the Maestro controller, but currently do not, and others who do not use Maestro, but use a simple wrapper or something more complex list ZooKeeper may have the same problems we have with ZK currently.

In fact a return code for an unexpected but intentional shutdown may not be sufficient; a program like ZooKeeper may not have access to the return code and it might be useful if there was a completely different signal _in addition to_ a differentiated return code, such as the presence of an "admin maintenance" file of some kind.  But like Mike, I think that's going in the wrong direction.

This commit is a simple change that recognizes the current state of affairs and provides a workaround which may not be needed in the long term, but is helpful now.  It may not be needed because at some point _all_ unexpected region terminations may be crashes, of other mechanisms for shutdown are removed. However, that is currently not the case, so this PR assists with the current gap.

kf6kjg:
@Appurist a restart command already exists and the relevant code to make that a straightforward shutdown with a unique exit code is in the diff in OpenSim.cs:780

appurist:
Ah yes, it's been so long since I reviewed this PR ... and you've already changed the (internal) restart operation to not call the console `shutdown` anymore, and differentiated the return codes from the two commands. If we update ZK to actually issue a `restart` when it's a restart, that will be even better.

From my perspective, there's no reason to hold the PR now.  The only reason I can think of would be if Mike (or someone) planned to immediately update the grid services and region server to Windows services (and Linux daemon/services), in which case the commands likely disappear completely.  

Even then, we'd need to change the implementation to notify the controller/manager to request a restart, which ties the grid server to a management interface (something new), which is something we have no plans to do at this time.

So this PR is now good to go as-is, from my perspective.

kf6kjg:
That new commit simply adds the default ext code of 1 to the other services, each of which already get the "shutdown" and "restart" exit codes because of using the same subsystem.

PR #241, opened 2016-07-05
Swagger and Postman documentation of the two remote access APIs

kf6kjg:
The XMLRPC spec was a real pain. They should be good and are at least much more condensed than the C# code they derive from.


kf6kjg:
I dub this PR complete as it stands, barring input from others.    It's nice and clean with no functional changes to the source code.

I'm preparing a different branch with some actual meaty changes to actually fix bugs I found while making this.

kf6kjg:
Well, turned out I'd missed the username and pass literals I'd left as temporary workarounds.

PR #242, opened 2016-07-06
Asset statistics ("show assets")

jimtarber:
- Adds a "show assets" command which dumps several stats related to asset reads and writes, plus the cache.


jimtarber:
We should probably discuss this one in more detail later today.  The removal of the _assetCache.HasAsset call in particular. While something along those lines needs to be done to resolve the problem, an alternative approach would be to provide an assetCache.Update to replace a wire asset with the corresponding stream and call that if HasAsset was true, or to split the StoreAsset into GetStream+StoreAsset and add the asset to the cache between those calls.

jimtarber:
I've reverted the "More CF asset handling changes" so as to split the stats commits from the logic changes. I branched to ```asset-logic-tweaks`` before reverting so that branch includes this reverted commit: 50da64c860c3f1630022e2bdcf78b5f8bfd8aa94  I've also restored some stats reporting changes lost in the revert, and doing one more commit to tune the report to be more readable and practical.

PR #243, opened 2016-07-11
Crossplatform remote access API authentication

kf6kjg:
Under Mono there's no way to check the username+password against the user, mainly because there's no guarantee that passwords are even used.

I realized that this system was acting more as an auth token than a user login. So I simplified the concept to a single passcode.
However, their's no guarantee that the configuration files or the system's memory will be kept secure.  So I decided to keep the code around only as a salted SHA256 hash.
I didn't go to the extreme and implement key stretching, but that could be added if there's enough worry.

Under .NET on Windows the old active user authentication process is still used if the hash is blank or not set.  This provides backwards compatibility with pre-existing setups.

### DIFF
--- a/InWorldz/InWorldz.RemoteAdmin/RemoteAdmin.cs
+++ b/InWorldz/InWorldz.RemoteAdmin/RemoteAdmin.cs
@@ -64,9 +64,12 @@ namespace InWorldz.RemoteAdmin
         private Dictionary<string, Dictionary<string, XmlMethodHandler>> m_commands = 
             new Dictionary<string, Dictionary<string, XmlMethodHandler>>();
 
+        private string m_remoteAccessHash = ""; // Empty string never matches any hashed value so is quite a safe default hash!
+        private string m_remoteAccessSalt = "";
+
         public delegate object XmlMethodHandler(IList args, IPEndPoint client);
 
-        public RemoteAdmin()
+        public RemoteAdmin(string access_hash, string access_salt)
         {
             AddCommand("session", "login_with_password", SessionLoginWithPassword);
             AddCommand("session", "logout", SessionLogout);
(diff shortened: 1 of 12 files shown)


mdickson:
This change will break existing Maestro code which uses the Util method to do authentication.  If we wanted cross platform might have been better to overload it there.

InWorldz/InWorldz.RemoteAdmin/RemoteAdmin.cs
```diff
@@ -189,14 +195,27 @@ private void sessionTimer_Elapsed(object sender, ElapsedEventArgs e)
...
+        {
+            return Util.CreateSaltedPasscodeHash(m_remoteAccessSalt, passcode);
+        }
+
+        private bool CheckRemoteAccessCode(string username, string password)
+        {
+            StringComparer comparer = StringComparer.OrdinalIgnoreCase;
+
+            // If a remote passcode was not set up then check with the system - however that only works on Windows under .NET.
+            return m_remoteAccessHash.Length > 0 ? 0 == comparer.Compare(CreateRemoteAccessHash(password), m_remoteAccessHash) : Util.AuthenticateAsSystemUser(username, password);
```
kf6kjg: This line does the backwards compat with the existing auth system that Maestro et al use.  If the remote access has is not set or is blank, the old Util method gets used.

kf6kjg:
I started out researching libpam auth methods. Then I found out why Mono doesn't already call it: PAM can use a large variety of authentication schemes: anything from simple username with or without password, to fingerprint, dongle, and more as people get inventive.  What's more, PAM only works under Linux, there'd have to another whole process for OSX.

Hence after reviewing all the options I decided that it was far simpler to just keep it in the code. Therefore I found a way to do so such that this code could go on IW or other servers without any config changes and work just like the old code did.

ddaeschler:
I'm creating a JWT token based auth that can use system usernames and check godlevels for API authentication. I should have something up in the next week or so for the remote console.

jimtarber:
Can this PR be closed now, in light of #271 #273 and #274 ? It remains an outstanding pull request (with conflicts now).

kf6kjg:
Ya, if any parts need resurrecting it'll be a totally different PR.

jimtarber:
Okay, it sounds like my oar-backup-tweaks branch, which was thought to be a good change but is on hold. Keep the branch around if you think there's value in there, but if there's no intention to merge this one into master, I'll close the PR.

PR #244, opened 2016-07-12
A small fix regarding setup on new installations

cb-dtf:
The logic should be untouched (besides the formatting) 
but on IOExceptions that happends on first-time-installation, it now creates a new Region settings file properly.

I am shutting down my fixes for now, so might as well share this little contribution.

### DIFF
--- a/OpenSim/Framework/Configuration/XML/XmlConfiguration.cs
+++ b/OpenSim/Framework/Configuration/XML/XmlConfiguration.cs
@@ -144,16 +144,33 @@ namespace OpenSim.Framework.Configuration.XML
             // Perform commit out of place in case of insufficient disk space.
             String oldFile = fileName + ".old";
             String newFile = fileName + ".new";
-            if (File.Exists(newFile))
-                File.Delete(newFile);
-            doc.Save(newFile);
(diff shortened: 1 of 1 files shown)


jimtarber:
So you were seeing an IOException on the first run? Is this because the region was running through the new region "wizard" prompts? It should really be committing those changes after getting the answers to the prompts so that this code doesn't have exceptions. That said, this looks to handle it.

However, the "file juggling" code I added recently to deal with low-space conditions on disks does assume the current ```region.xml``` file exists. Rather than handling the exception, it seems we could probably avoid it by changing the ```File.Move(fileName, oldFile);``` to include the same conditional that is present in the ```catch```. That is, leaving the exception handler in place but also avoiding it in the first place.  If Dave merges this in, I'll probably add the same conditional to the two ```Move``` calls (if you haven't).

But regardless, this PR looks good. Thumbs up and thanks for noticing and fixing it. It's not something we'd see on the InWorldz regions or in testing.

cb-dtf:
Thank you for the nice words.

To your question, yes I only saw the Exception on the first run, so I guess you are right about it would be even better to make sure it doesn't happen in the first place.

PR #253, opened 2016-07-21
Fixed llGetObjectDetails(id, [OBJECT_OWNER]) to return NULL_KEY for deeded objects

jimtarber:
Fixed llGetObjectDetails(id, [OBJECT_OWNER]) to return NULL_KEY for group-deeded objects, as per SL wiki and tests in SL to confirm it.


jimtarber:
Ignore the name of the branch, it's OBJECT_OWNER that was fixed.

PR #255, opened 2016-07-31
Added support for IP bans in the User login server (rather than region)

jimtarber:
Supports dynamic updates to a bans.txt file which is a list of IP address segments (e.g. partials) not permitted to log in.

Sample bans.txt (place in the bin folder of the User grid service handling logins):
```
; This is a comment because it will never match an IP address
; The first will match the full 123.45.67.* subnet, the second is specific.
123.45.67.
98.76.54.123
```


ddaeschler:
Looks good

PR #256, opened 2016-08-02
Login bans & other user errors treated as region down, causing problems.

jimtarber:
This meant if a banned user tried to log in a lot, they could deny direct logins to that region by ***all*** users, again and again. See http://inworldz.com/forums/viewtopic.php?p=208345#p208345 

Fixed any good response from the region (including user fails) to not be interpreted as a region *down* failure, blocking valid users after 10 errors in 5 minutes. Also renamed the booleans to be clearer, and fixed the return code to use both.


jimtarber:
We might need to update the User server as this is causing problems on IDI and on the "Little & Big" region, perhaps others.  The symptom is that the region appears down on a direct login from the Login screen using the region name.

PR #257, opened 2016-08-03
Map tile correction and optimize (2x speedup)

kf6kjg:
Draws the objects correctly as far as it goes: the colors used are per-face instead of per-object, they are NOT inverted, and OBBs are scaled correctly, not 2X.

Before any optimization:

```
06:13:23 - [MAPTILE]: Generating Maptile Step 2: Done in 920 ms, Filtering/Prepping took 514, Array cloning took 1, Sorting took 15, Drawing took 390 (=1840ms)
06:14:24 - [MAPTILE]: Generating Maptile Step 2: Done in 839 ms, Filtering/Prepping took 433, Array cloning took 0, Sorting took 15, Drawing took 391 (=1678ms)
06:15:26 - [MAPTILE]: Generating Maptile Step 2: Done in 1315 ms, Filtering/Prepping took 468, Array cloning took 453, Sorting took 18, Drawing took 376 (=2,630ms)
```

After this patch (with the metrics uncommented of course)

```
05:56:38 - [MAPTILE]: Generating Maptile Step 2 (Objects): Processed 44743 entities, 44992 prims, 44727 used for map drawing, resulting in 44740 faces to draw.
05:56:38 - [MAPTILE]: Generating Maptile Step 2 (Objects): Timing: prepping took 6ms, filtering prims took 87ms, calculating vertices took 66ms, computing sorting height took 14ms, calculating OBB normal took 31ms, getting OBB face colors took 114ms, calculating OBBs took 17ms, adding OBBs to list took 10ms, sorting took 78ms, drawing took 94ms, total time: 745ms
05:56:50 - [MAPTILE]: Generating Maptile Step 2 (Objects): Processed 44743 entities, 44992 prims, 44727 used for map drawing, resulting in 44740 faces to draw.
05:56:50 - [MAPTILE]: Generating Maptile Step 2 (Objects): Timing: prepping took 6ms, filtering prims took 94ms, calculating vertices took 62ms, computing sorting height took 18ms, calculating OBB normal took 26ms, getting OBB face colors took 98ms, calculating OBBs took 24ms, adding OBBs to list took 5ms, sorting took 87ms, drawing took 94ms, total time: 775ms
05:57:02 - [MAPTILE]: Generating Maptile Step 2 (Objects): Processed 44743 entities, 44992 prims, 44727 used for map drawing, resulting in 44740 faces to draw.
05:57:02 - [MAPTILE]: Generating Maptile Step 2 (Objects): Timing: prepping took 6ms, filtering prims took 83ms, calculating vertices took 62ms, computing sorting height took 14ms, calculating OBB normal took 41ms, getting OBB face colors took 99ms, calculating OBBs took 22ms, adding OBBs to list took 11ms, sorting took 83ms, drawing took 94ms, total time: 761ms
```


appurist:
Just reviewing some of these older PRs again.  The changes for these tile optimizations look good, except I'd feel more at ease about them if the timer wasn't restarted _before_ the processing started, but instead was restarted after it completed.  WorldMapModule.cs line 1173:
https://github.com/InWorldz/halcyon/pull/257/files#diff-a9e3bf7a3b06e734944bda023851354cR1173

I only have one slight concern, and that is that the minimum time should be the minimum _between_ runs, not the minimum between start times. i.e. excluding processing time.  If it's set to 60-second refresh, and it takes 20 seconds to process, there should still be 60 seconds of idle time between those, not 40 seconds (the difference).  Just in case something goes wrong somewhere, and it ends up taking a minute (or more) and it's set to refresh every minute (that would be a tight loop then). If it's the minimum time between, it guarantees at least that much idle time.

kf6kjg:
The big question is whether or not the entire operation should be removed from the simulator: separation of concerns etc.  This section of code is only for generating map tiles, not even the minimap uses this.

appurist:
That doesn't actually address my concern, which admittedly was a bit academic. Mostly that the minimum wait time is not really a minimum _wait_ time, and I know the usual processing time is measured in seconds.  It's just that by moving the `timer.Start` a couple of lines later, it avoids all of this, and provides for scalability beyond the current map sizes.  (Just ask yourself if this handles 100 times the data. No, but it would with a one-line move.  And even with 10 times the data, it would be using about 50% of a processor, plus or minus a lot since we don't really have a good handle on how long a run will take.)  I was just pointing out that these considerations were moot if the minimum wait time was the time between the _end_ of processing and the start of the next run.

kf6kjg:
Ok, so maybe I should have typed "semi-address" :P

I agree that doing so would address the rendering phase delay scaling issue your bring up.

kf6kjg:
While Anaximander 2 generates its own map tiles now, the Halcyon server still has the capacity to do so, so I'm leaving this open until such a time as its either merged or it's decided that the map tile generation subsystem is to be removed.

appurist:
I'm going to merge this for now but if Anaximander 2 makes this obsolete, I won't shed any tears. ;)

PR #260, opened 2016-08-05
Added a "teleport user" console command.

jimtarber:
Allows forced-teleport of an avatar to another region (or same region).
```teleport user FirstName LastName RegionName x y z```


jimtarber:
This should be safe to merge as it is all new code only triggered from the console.

PR #263, opened 2016-08-06
Change ObjectGiveMoney to disallow payments of zero

zauberparacelsus:
This makes llGiveMoney compatible with its Second Life implementation.


jimtarber:
I misread this commit at first, thinking it was allowing $0 transactions when in fact it's disallowing the money transfer of $0, skipping the transfer (as per function documentation).  There are two semantic changes by this:
- A script that checks the return code will get a false/0 back now for $0 when it used to get true/1.
- The notification of "You were paid Iz$ by user" is now skipped as well.
This will reduce messages and code executed.

This change should be completely safe although the first point above is the largest risk. However, since it _always_ returns that same false/0 in SL, I don't think anyone would be checking the result, and certainly not aborting important steps that followed, or the script wouldn't work at all in SL.

So I think we can proceed with this one. Ready to merge.

PR #264, opened 2016-08-11
Avatar-as-a-prim

jimtarber:
Major changes for avatar-as-a-prim (AaaP). Should be mostly backwards-compatible with existing scripts where possible. If a script positions or rotates an avatar as if it was a prim, that avatar (SP) switches into avatar-as-a-prim mode (AvatarMovesWithPart becomes false). This AaaP mode resets to legacy mode if the avatar stands or leaves the region.

This PR may also fix obscure problems with culling missing or including some prims inappropriately, as well as several timing problems on region crossings/entry that could place the avatar in the sky or at 0,0,0.


jimtarber:
Hold this PR for now. I'm going to do additional sit target-related changes.

PR #270, opened 2016-09-09
Region confirms crossing

jimtarber:
These are the changes to allow the receiving region to directly confirm a crossing success rather than waiting for the viewer to do so with a CompleteMovement message.  The ban list checks were already doing server-side destination confirmations and replying with an success/fail indicator.  After this change, if it okays the crossing, it's good; the agent or object has crossed, and we can consider the timeout on the sending side answered.  What remains is just a matter of the viewer notifying the region when it _too_ is ready.  But in the meantime, any prim or agent is considered there by the sending side.  This really helps in some cases, especially single-vehicle flights, and probably even more so if there are more passengers. It eliminates Internet latency from the crossing, and things like the viewer purge can't cause a crossing fail. Also, it means there won't be cases of two root agents on timeouts anymore, since timeouts have been eliminated.

The main change is in https://github.com/InWorldz/halcyon/commit/1a85f3261f042f285d3183d38b16f6bc8311018a where the call to SendReleaseAgent is factored out and now also called asynchrnously after the destination region authorizes the crossing.

This PR also adds timers around the viewer confirmations, both for the seated avatar and for the CompleteMovement, as well as a viewer-side server chat command to have the regions report the timings to the viewer: ```/!debug crossings```


jimtarber:
The merge conflicts with #269 changes have been resolved.  Ready for merge to master.

ddaeschler:
I want to merge this but my concern is releasing the agent before the viewer has a connection orphans the session entirely when the viewer fails to complete the handoff. I don't know what this will do from a practical standpoint though.

Also, what would happen in the case of a teleport far away? Wouldn't this close the circuit on the sending side before the viewer actually completes a connection to the new region? I suspect you'll get a lot of teleport disconnections if so.

jimtarber:
The teleport case is actually a huge improvement.  In the past, if the viewer took too long to respond, the user was left with a _root_ agent in _both_ regions. (Actually that also applies to the crossing case too.)  This resulted in major agent ownership problems that required a relog in order to correct.

Now the viewer must still check in at the destination region in order to get things like controls, but the normal connection timeout applies rather than the much shorter crossing/teleport hand-off timeout.  A crossing or teleport is an agreement between regions over who is managing the session.  In the past, that required a "vote" from the viewer which could effectively veto a crossing, but without informing the destination region that it was vetoed, thus resulting in ownership confusion.  Also during region change, the transition time where no region really owned the session was much larger, resulting in a period where the viewer (even if it agreed) could not control the vehicle.  In all cases the teleports and crossings should be much faster now, and some operations occur much earlier, reducing or completely eliminating awkward race conditions and timing windows (e.g. the ContinueSitAsRootAgent method is called ahead of the viewer confirmation and occurs _much_ sooner now, resulting in proper parentage being established immediately).  It also means slow passengers no longer hold up passengers/drivers with faster connections.

The only problem I have encountered so far is that the SP.RemotePresences.HasConnectionsEstablishing() check that results in the "An avatar can not begin transition to a new region while connections are being established to neighbor regions" exception being thrown can go off on a corner crossing between 3 regions (this occurred once for me in testing), and it is unclear to me why the pre-check for the object crossing did not just prevent the second crossing attempt.  It also pre-checks the same SP.RemotePresences.HasConnectionsEstablishing() check.  Unless they do not _default_ to establishing so there is a race condition where the precheck passes but Establishing() becomes true for the actual avatar crossing.

jimtarber:
The HasConnectionsEstablishing() precheck I'm referring to is in SP.TestForAndBeginCrossing()... 

The HasConnectionsEstablishing() that fails and throws is in TransitBeginState.StateEntry().

ddaeschler:
That error doesn't sound like a new condition. Is it simply easier to trigger now without waiting for the viewer vote?

jimtarber:
That condition definitely isn't new, but I think it just occurred to me what the problem is. Since the region considers the agent to be in the region much faster now, the object is free to _move on_ sooner. So yes, it's much easier to reproduce the condition now with a moving object that continues to coast (or fly) after the first crossing resumes operation.  So I guess I'm a bit less worried about it in the technical sense, but I am still concerned we may see more cases of it in the practical sense.  And the problem seems to be that we aren't handling it well. The recovery seems to bust the session. (I relogged when it happened to me). Although I found the plane I was flying had (eventually) successfully completed the second crossing, so it probably retried again after my viewer confirmed the first crossing.

So I think this is something I need to follow up on and try to determine if our error handling in this case needs an additional tweak. But I think you're correct in that it's probably not a new case, just perhaps easier to reproduce now.  So I think we're good to proceed in this case, except that there may be an additional (existing) loose end to work on that may matter more now.

ddaeschler:
one thing to consider is that if this allows the user to get to the edge of the new region before they're actually connected to it, the viewer will never get our requests to connect to the subsequent 3rd region in a high speed run. This was one of the reasons for the wait. I would check to see what happens in this case. A super fast crossing vehicle with this new code

jimtarber:
The viewer should already be connected (as a child agent) in any crossing.  It's not the connection that is happening here, it's the CompleteMovement packet.  We were always making it the root agent where we are now, but the sending side could time out on its downgrading to child agent.  Once the destination has confirmed it can (and intends to immediately) call MakeRootAgent, we don't really have any choice but to confirm that to the sending side.  Failing to do so is why we end up with dual root agents on a viewer timeout.

In terms of the object reaching the far side, that's easy enough to do with a normal vehicle on a corner crossing (via a brief transition through an intermediate region).  I'm not sure why the rollback in that situation (which isn't new) isn't going well, at last in my "occurred once" failure. It could be that the object is allowed to continue on, because it has all seated avatars, even though the state machine for the crossing hasn't yet completed out back to normal (non-transitioning) avatar yet.  Perhaps the avatar transit state machine is still waiting for the CompleteMovement packet before it considers the crossing complete.  Hmm. That made me think of something: I was thinking that the RemotePresences.HasConnectionsEstablishing() precheck was preventing exit from the region pending complete entry, but I don't think it is. I think I missed the RemotePresences part.  It's not checking that the agent is ready in _this_ region, but that the viewer should have a connection with the destination region (before allowing it to enter that region).  So perhaps what is missing here is a "avatar is successfully in this region" before allowing it to leave again.  Yes... it will already have a remote presence on the destination region on a corner crossing. And with a >256 draw distance, it will even have one to the region on the far side _before_ entering the intermediate region.  So we probably need to confirm it's in this region before allowing it to exit.  I don't think this is a new problem, but the timing window may have grown by an order of magnitude here.

ddaeschler:
This sounds good, and is thoroughly researched. I'm good with the PR as is unless you want to add more first.

jimtarber:
I spent a few more hours last night investigating the corner crossings situation. Flying across was actually handled properly except that the "bounce-back" code was a fixed unconditional reversing of the direction, which typically worked (although was a bit heavy in bounce-back at times).  We may want to use that approach in the future but with an absolute max bounce-back.  It was great at small velocities, but at speed the jump was a bit jarring.

At any rate, the real issue there was that on diagonal/corner crossings of avatars flying from A->B->C, the bounce-back from C (while in B) could actually cause the bounced position to be inside A again.  I'm not sure exactly what the problem was there, but I just made it conditional on the new position being within the region and it's actually really nice now.  

The old bounce-back is retained for the normal cases, although as I mentioned above, I think it could use an upper limit on bounce distance, such as 0.5m.  I'm crossing by flying into a corner and ending up at 255,250 or even farther (I didn't actually check the position, only that I was way down the edge of the region, well off my original path which came from the other region).

All of the above only applies to the flying (or walking) avatar case (although typically flying in order to do the second crossing attempt too soon).

jimtarber:
The seated corner crossing case is a bit more complex.  
HasConnectionsEstablishing
The problem seems to be that the corner crossing attempt throws the exception, not because the connection to the new region is being established, but because another piece of code is active so the semaphore count is 0.  (It's not getting to the foreach to check the actual remote presences, which are both established.)

I'm not sure that semaphore count check is appropriate there. Or perhaps the fix is is just that that conditional _should be inside the lock_ so that the caller waits very briefly to gain access to the remote presences list.  It's not appropriate to throw an exception just because some other code is looking at the remote presences.

jimtarber:
I should also add that I'm trying to determine what the other thread is doing that is causing the semaphore to be in use. I think it is probably a CalculateAndResyncNeighbors when the avatar enters the middle region  A -> B (resync) -> C.

jimtarber:
Also, I haven't followed through with the merge because I feel we're close to taking care of the corner issues, which may have been longer-standing issues just made easier to debug now because the timing window is wider now.  I'm hoping to wrap things up tonight though, do the merge, and build an RC.

mdickson:
Does this approach still work with teleports where there is like no child agent on the remote?  Is it possible we'll see more fatal teleport failures (as in if the teleport fails the client disconnects)?

jimtarber:
Yes, it works with no child agent on the destination (teleports). In fact it auto-repairs that for the neighbor cases, reestablishing the connection.

This change should result in far fewer teleport failures, eliminating the viewer response delays from being a factor.  Many error cases have been eliminated, most notably the timeout waiting for the viewer, which is ye olde "child agent update failed" error we used to see all the time.

jimtarber:
This discovery with HasConnectionsEstablishing may explain several of the longer crossings on the main grid, as well as crossing failures.  The problem is that it is checking for **any** connection establishing, among the 25 regions in the 5x5. In the case of an actual (neighbor) crossing, there should already be an established (child agent) connection there, but if there are more (new) neighbors on the far side of the destination region, this check will require those to also be in place before allowing the crossing into the region that already has the established connection.  Furthermore this timing problem with the semaphore being in use allows the possibility the simple crossing into an immediate neighbor (with the established connection) to fail.

I'm testing a change now where that test is replaced by a new HasEstablishedConnection(destRegion) check instead. We don't really care about the others, or if the semaphore is in use, if there's an established connection. We just want to keep going without delay.

ddaeschler:
Having a consistent state before moving on is VITAL to making sure connections are properly cleaned up. You can't just move on with connections in transit or the viewer will be left connected to smattering a of regions as you move forward.

jimtarber:
True, however in the past the agent may have been allowed to move on prematurely, because it only checked the avatar connection state (for example, it did not verify that the agent was a root agent in the current region). This is why I think we had trouble on the corner crossings after my change, until I added the IsFullyInRegion check to verify we had the CompleteMovement message from the viewer.  A connection is not the same as a fully-ready root agent.

But I think your point is that the set of 25 regions changes on each crossing, and therefore allowing the object or agent to cross with only the immediate neighbors in place (or in fact the current and destination regions) may be premature.  I don't think the viewer is getting anything different here, it's just that there is the equivalent of a sliding window (more asynchrony) because the IsFullyInRegion is triggering the messages to the viewer for all (up to) 25 regions.  But also once the viewer does arrive in the third region, as soon as it transitions the agent to root agent, that region will also call CalculateAndResyncNeighbors based on the final destination.  Yes the handcuffs holding the avatar or object in the region have been removed, but the only regions that truly matter in this case are the ones that are still checked.

ddaeschler:
We need to have a consistent view of all the regions that we have requested the viewer to connect to and know whether or not that connection has been established because it is our job to tell the viewer to tear down those child connections when the regions go out of range. If any subsequent commits break that functionality or cause inconsistency in our view (eg we sent a connect but didn't wait for the ack and didn't consider that region connected to even when viewer connects later) it will cause major problems.

The reason I'm making that comment is the semaphore is decremented to zero only after all region connection requests have completed or failed and that is vital to keep in place so our view of what regions the viewer has connections to matches the list in the viewer. That's why the message happens

jimtarber:
The region holding the semaphore won't be controlling the viewer a few milliseconds later, but I don't see why it wouldn't have sent the disconnects while it was.  We wait for the CompleteMovement in the current region before allowing the exit, and recalc the neighbors.  But if I follow the comment above, I think the case you're worried about is the corner crossing (or super fast crossing) where we are going to send the disconnects but one or more of the 5 might get dropped by the network. In the event that such a fast transition happens (e.g. corner crossing), if this happens, I didn't actually think the disconnects were absolutely required; that instead the viewer connection to the region would disconnect on a time out (or because it is out of range). The most important thing is that the new 25 all be successfully added, and that is handled by the last region in the avatar's travels.

I really think that much of the trouble we see now is in some way, probably indirectly, related to the increased neighbor range. Moving to 5x5 instead of 3x3 means that there are up to 10 connection updates per crossing rather than the previous maximum of 6, although I think they are all sent asynchronously, and thus in parallel.  So there are 5 disconnects on the main grid, often to heavy content regions, even in the case of traveling the straits.  In the usual case of 1 crossing only, the avatar isn't going to be moving on anyway.  And in the corner crossing case, the important thing is to establish the new connections and complete the crossing as quickly as possible, so as minimize the time the avatar is in transition.  I think you're correct in that this could, given network unreliability, introduce cases where the viewer is temporarily attached to more than 25 regions, but I am pretty sure those connections will be closed by the viewer after some time.  It has these connections normally anyway, on a crossing, it's just that we're talking about a longer period.  Here is a much earlier crossing from I'z Straits 42 to 42:
Crossing 42 to 41
This is the case where there were 5 void regions in both 5x5's, so it was normally 20 neighbors.  It temporarily expanded to 25 regions as it was a 6x5 with 5 voids, then down to 20 again.  I think the crossing failed in this case because of one or two of the regions being disconnected were slow. (The avatar in question was the EO on two of the regions well behind her being disconnected.)

So what I'm trying to do here is to eliminate the cases that are causing the most trouble, and most of that is waiting for the round trip time for the viewer to process 5 new connections and 5 new disconnects (and 5 viewer purges).  When in fact it is already connected to the source and destination regions in question, and the crossing can be handled by the region two servers directly.  Anything else that happens needs to happen asynchronously.

jimtarber:
I guess the real question is how you think I should proceed. What I was planning on doing was to create an updated RC from this branch and anything else that gets merged into master, and test this on the Sandbox regions and if that goes well, the Straits and Ocean regions.  I can monitor the list of connected regions (e.g. like the top pane of the Message Log, as in the image in my previous comment) and see if we're leaking connections on a fast double/corner crossing.  

If there is any problem there, it seems to me that the destination region knows the previous region and can therefore (redundantly) notify the viewer of those disconnects. In fact, it could be argued that it would be a much _better_ design for (only) the destination region to send the disconnects to the viewer, after the crossing has completed. (That also pretty much completely avoids the "viewer purge" issue's impact on crossings.)

ddaeschler:
I'm going to have to do a code review on the branch and try to figure out if there are risks. I don't want I keep adding changes because it gets really difficult to know where something has gone wrong then and crossing is a sensitive issue for the community

ddaeschler:
I should be able to do a thorough review on Monday. Let me know your availability on our normal channels and we'll review

jimtarber:
Monday is good (especially afternoon).

However in terms of this release, I'm thinking at this point I should build an RC _without_ crossing changes (my original 0.9.28 plan) and we can do another soak test with just crossing changes as soon as we have something.  We can do a .29 RC with the crossing changes and update the Sandbox regions and possibly the Straits and Oceans at any time for a soak test.  In fact if we stuck to the schedule, that would give us a 3-week soak (although we may be able to gain good confidence faster than that and do a quick "crossings" release).

jimtarber:
Interesting. I built a release candidate that does not have any of these crossing changes at all https://github.com/InWorldz/halcyon/tree/release-0.9.28 and did some basic tests in the updated Sandbox regions. My first attempt at a corner crossing ended very poorly, basically failing in the same manner that I found during my testing:
Sandbox test
So the problem wasn't new in this branch, and is in fact now fixed in this branch.  

I think this branch has some significant fixes and improvements; the question is at what (reliability) cost.

We'll review this Monday, but quite frankly, there are known serious problems in the current code, and the largest problem that we suspect might be present in the updated branch is that it may leave a connection open longer than is necessary (or even indefinitely) when leaving a region.

ddaeschler:
I don't like any solution that leaves known problems because they'll just need to be solved later.  in fact, I don't understand with that error how the crossing even took place enough to allow that half in half out condition as it should've aborted the crossing before it even began as part of the original design.

We'll talk Monday

ddaeschler:
Definitey build the RC without the crossing changes. As history has shown that is usually the best way to go with anything crossing. There is not rush on this. I'd rather know we put the extra time into handling the edge cases

jimtarber:
In this case of crossings, I agree we don't need to rush any changes into production.  Changes of this nature probably need the support of both of us and it didn't have support from either of us on Thursday evening when I built the RC _without_ the crossing changes. I've already announced they are postponed until 0.9.29 (at the earliest).

But I disagree somewhat with your wording: while I agree with not liking any solution that leaves known problems, that's exactly what we're doing when we postpone a fix-in-hand that leaves a _worse_ situation in the production environment. I think the approach to-date has been one of incremental improvements, and I think that should continue. Each release should be better than the previous.

I don't have a concern with postponing the crossing fixes (in fact that was my original plan, until a small fix looked promising), but I think it should be a matter of comparing one set of known problems versus the risk of new code and any known problems in that, and appraising the confidence level in the changes.  In general, I am not comfortable with withholding changes that appear to be an improvement, just because they may require another followup in order to resolve one remaining _lesser_ issue.  I think in cases like that, the impact of the _current_ production issues needs to be weighed against the impact of that remaining known issue in any update.

But this crossing process is tricky to get right and the presence of one small oversight can cause all hell to break loose.  That said, we've done some pretty heavy real-world testing of the initial change, found it quite an improvement but still saw some cases that made me look deeper.  I think I've identified some of those now, however, and that is the nature of this followup to the PR.

I think it's generally good but we need to do additional functional testing to measure the outstanding region connections after crossings.  My main concern Thursday night was that we didn't have a clear understanding of the impact on past region disconnects, so additional testing is needed. But it was an interesting thought to just abandon the disconnects performed from the source region entirely, if the avatar starts to move again into yet another region, and to let the final destination region handle _all of_ those disconnects.  In addition to tackling that outstanding issue, it may also be a sizable decrease in the time to complete a crossing.

jimtarber:
_"I don't understand with that error how the crossing even took place enough to allow that half in half out condition as it should've aborted the crossing before it even began as part of the original design."_

I think there's a quick answer to that: it allowed the object to cross because the semaphore was not in use at the time of the object check, but then it was in use at the time of the avatar crossing so the avatar crossing was rejected (even though the avatar was seated).  My change in this PR ensures that these checks and conditions are performed up front before allowing the object to leave the region.

ddaeschler:
That's simple. The production environment code has been in place for years and has whatever issue triggered this PR for as long. I don't like swapping out known issues for unknown issues just because something has all of a sudden been deemed "critical" by someone. I've seen that make things worse far too many times when the unknowns hit Harder than expected.

I'm sure the new code is better. I simply want to explore the new issues that may be raised before giving it a go ahead to understand the full scope of their possible impact.

jimtarber:
That sounds good to me, and the plan to include this in the followup release (only if ready then), not the current one, is already [in motion](http://inworldz.com/forums/viewtopic.php?p=210482#p210482).

I don't really think there's a sudden urgency here, as much as I've been tied up with other things (like avatar-as-a-prim) and as each of those clears, I've looked at the outstanding issues and tried to find the most critical to fix next.  What is sudden is that the avatar-as-a-prim work seems to be complete, so I'm taking a very close look at crossings.  The state of crossings is not what I thought it was for the last year, and I don't like to move forward without a stable foundation, and that includes reliable crossings, so this one made it to the top of my list.

I thought it was going to be a small change that could be optionally included, but the concern you raised leaves questions unanswered so I definitely agree we should understand the impact of that more before deployment.  

I do really think this is an urgent problem though. The fact that most users have been mostly patient with it for a year doesn't mean it's less urgent.  (I'll send you one of the messages from the vehicles groups that is accurate and shows the impact on the community.)  I think we are losing users due to this problem and we need to address it as a priority. The good news is that it _is_ a priority for us, we do have some relief queued for testing, and hopefully we'll have taken care of it (or at least improved things) with the 0.9.29 release.

jimtarber:
I have restored the full region connection tests before leaving the region.  There are actually more there now that I had already added to fix the corner crossings case (e.g. IsFullyInRegion) but the key point of the last commit is that the concern regarding moving on too soon was well-founded and I concur that is not an option.

On further discussions with @ddaeschler, and after considering alternatives (destination region sending enable/disable for whole chain of crossings), we cannot allow an avatar to initiate a new crossing while connections are being established. The chronological order of region connection adds/removes must be preserved, especially given the possibility of reentry into a region, and we cannot ensure that, once an avatar leaves the region.  It's a shame that we need to block and wait for more than just the destination region, but it is what it is, the alternatives are worse.

I think this one is ready for merge again now, and RC testing.

jimtarber:
Testing of this update found that it did not resolve the "crossing case", or any other case where an avatar was sitting on an object that attempted to leave a region before the avatar had finished entering it.  The problem here was basically that the avatar.RemotePresences.HasConnectionsEstablishing() checks were done at the avatar level, i.e. _after_ the prim was allowed to cross into the destination region.  The latest commit now includes a precheck when the prim is attempting to leave the region that prevents departure until any seated avatars are ready to cross.

ddaeschler:
I should be able to review this final version f the PR today.

PR #271, opened 2016-09-11
Jwt remote console

kf6kjg:
Could do with some more hardening, but it is fully functional.

### DIFF
--- a/InWorldz/JWT/JWTSignatureException.cs
+++ b/InWorldz/JWT/JWTSignatureException.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace OpenSim.Framework.Communications.JWT
+namespace InWorldz.JWT
 {
     public class JWTSignatureException : Exception
     {
--- a/InWorldz/JWT/JWTSignatureFailureCauses.cs
+++ b/InWorldz/JWT/JWTSignatureFailureCauses.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace OpenSim.Framework.Communications.JWT
+namespace InWorldz.JWT
 {
     public enum JWTSignatureFailureCauses
     {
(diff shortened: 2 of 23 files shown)


jimtarber:
Some fundamental info missing here. What is the overall purpose of this PR? To add JWT authentication, only to the existing RemoteConsole support? So is this accessible from a viewer? From a different client such as a webapp? This runs as part of the region server executable?

ddaeschler:
This is a MOSES requirement for an HTML 5 viewer.

ddaeschler:
What's up with the diffs against OpenSim/Region/Framework/Scenes/ScenePresence.cs and the search module?

kf6kjg:
Agreed, and I apologize for the scant info.

The purpose of this is to allow cross-platform authentication in the manner that David was mentioning in PR #243. In fact, this downright supersedes PR #243 WRT RemoteConsole.  The Remote Access subsystem has not been matched to the same.

There are some wishes in the code comments: it'd be nice if the various services could be either registrable or be settable by an admin.  There's some merit to the thought that authentication and authorization should be separated.  This thought will need some review.

kf6kjg:
@David: Came from a merge from master I did at the beginning.  You might want to diff against master instead of the branch.

jimtarber:
@ddaeschler The weirdness of the diffs is that @kf6kjg has updated from master but I think the destination branch here hasn't. If you update from master, hopefully that will eliminate all the "extra" diffs.

kf6kjg:
You could also hit up a targeted diff such as https://github.com/InWorldz/halcyon/compare/6e057d9a12bd9bd93fe9eb731037c4757e73c2aa...kf6kjg:jwt-remote-console

6e057d9a12bd9bd93fe9eb731037c4757e73c2aa was the tip of master at the time I merged it in. Still is at this time.

ddaeschler:
Reviewed. Looks good. Thanks!

PR #275, opened 2016-09-15
Increased outfit name size limit from 36 to 255 characters

zauberparacelsus:
Matthew Beardmore says that the length of 36 was a copy & paste error on his part, while 255 is a length suggested by Jim Tarber.

This change will probably only affect new installations of Halcyon (or maybe not, I don't know the specifics).  Updating InWorldz to take advantage of this will require a database migration.

### DIFF
--- a/bin/inworldz-core-base.sql
+++ b/bin/inworldz-core-base.sql
@@ -212,7 +212,7 @@ DROP TABLE IF EXISTS `botappearance`;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `botappearance` (
   `Owner` char(36) NOT NULL,
-  `OutfitName` varchar(36) NOT NULL,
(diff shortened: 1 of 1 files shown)


jimtarber:
I don't see any problems with this, especially given the fields in question are varchar, not exceeding 255.  I don't think that causes any actual size changes in the db (still single-byte length), so it would probably be fast anyway.

For the InWorldz grid, @ddaeschler would need to apply the same kind of change with Percona tools (pt-online-schema-change).  (I could probably apply the schema change on the beta grid directly, but if we need to do it on the main grid, I'd leave that for Dave to take care of both at the same time. He can use the beta grid as a test run for the main grid.)

However, for new table creates, this PR looks good to me and can be merged.

ddaeschler:
Awesome. Thanks!

PR #276, opened 2016-09-15
Avatar as a prim part2

jimtarber:
I had a major portion of functionality missing from the _pattern_ used for each prim param operation, so this PR fixes that (the same simple fix repeated for every value of PRIM_...).

Also, the list of links didn't include avatars when LINK_* "wildcards" were supplied (e.g. LINK_SET) and again were not included when LINK_TARGET was used with a wildcard.  You had to explicitly provide a specific avatar link number. This has been fixed.

Also, the order of the links returned is now sorted according to link number order. Fixes Mantis 569: http://bugs.inworldz.com/mantis/view.php?id=569

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
@@ -10773,10 +10773,7 @@ namespace InWorldz.Phlox.Engine
             ScenePresence avatar = null;
 
             // Support avatar-as-a-prim link number.
-            if (linknumber > m_host.ParentGroup.PartCount)
-                avatar = m_host.ParentGroup.GetSeatedAvatarByLink(linknumber);
-            else
-                parts = GetLinkPrimsOnly(linknumber);
+            parts = GetLinkParts(linknumber, true);
 
             while (idx < rules.Length)
             {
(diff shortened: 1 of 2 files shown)


jimtarber:
This PR is actually fixes for an incomplete commit in #269 already merged which I have detected regressions for in testing.  But it's the same pattern fix repeated for all cases in the switch (of which there are many):
Code pattern
The parts list includes avatars now, and if you specified a wildcard like LINK_SET instead of avatar link number, it causes type conflict exceptions when it reaches the avatar. The places where the list is iterated need to differentiate the two cases. There are cleaner ways to do this but the ones I considered would require more significant changes.

jimtarber:
The other change in this (the second commit) is to return the parts list ordered by link number so that callers of script functions (e.g. LINK_SET, [PRIM_NAME]) know which links the names correspond to.  Resolves Mantis 569.  We'll want that behavior in other places anyway.

mdickson:
I'll have to take a look at the Materials code.  It  walks a list of parts and may need a similar check for type if the parts list can contain "avatars" as well.

jimtarber:
The internal calls to SOG.GetParts are unaffected. They still return only the prim links, no avatars.

ddaeschler:
Agreed that it doesn't look super pretty, but it'll do for now. Merging!

PR #282, opened 2016-09-29
Release 0.9.28

jimtarber:
Some small but important post-rollout fixes for issues with new code and regressions. Missing permissions check on llUnSit calls, StandUp applying to the root prim after use of avatar-as-a-prim features.


jimtarber:
I may have one more commit for this PR. The last commit may have been on case of a pattern that needs further resolution.  I'll know later tonight.

jimtarber:
Okay this PR looks good as-is. The remaining pattern cases will be fixed in a separate update because they do not result in code/semantic differences.  For example, some check the parent group (which is the same for child and root prims). Some only check for null, so all objects sat upon are equivalent for that test.  I'll fix it to be more technically correct in a separate update. I don't want to include any last-minute updates in 0.9.28 unless they are absolutely necessary/urgent for this release.

PR #286, opened 2016-10-06
Fixed missing Initialize call for loading load asset XML files.

jimtarber:
If config isn't passed in, _settings is null and LoadDefaultAssets() is not called to load the XML files from the assets subfolder.  I believe this is why the "missing" asset counters are so high.

### DIFF
--- a/InWorldz/InWorldz.Data.Assets.Stratus/StratusAssetClient.cs
+++ b/InWorldz/InWorldz.Data.Assets.Stratus/StratusAssetClient.cs
@@ -120,6 +120,7 @@ namespace InWorldz.Data.Assets.Stratus
                 if (Config.Settings.Instance.LegacySupport)
                 {
                     _whipAssetClient = new Whip.Client.AssetClient(Config.Settings.Instance.WhipURL);
+                    _whipAssetClient.Initialize(settings);
 
                     if (Config.Settings.Instance.CFSupport)
                     {


ddaeschler:
I think this was removed to speed up region startup times on the IW grid. We dont want to try to reload all those files on every startup. There is a tool to do this specifically as part of the stratus files

jimtarber:
The problem I have is that the existing code does not function. If we don't want it to load the assets, rather than _not_ passing the config object to the WHIP code, we should pass the config object with an option to disable the loading, or simply default the code itself to load not loading the assets and let that be overridden in the config options.  I don't see _not_ passing the config to this module as any kind of appropriate solution, because it also skips all of Initialize (e.g. SetupConnections is not called). I think might be having an effect on the fallback to WHIP assets.
```
        public void Initialize(ConfigSettings settings)
        {
            _settings = settings;

            Nini.Config.IConfig netConfig = settings.SettingsFile["Network"];
            if (netConfig != null && netConfig.Contains("asset_server_url"))
            {
                string url = netConfig.GetString("asset_server_url");
                this.SetupConnections(url);
            }
            else
            {
                throw new Exception("Network/asset_server_url is a required setting");
            }
        }
```
Perhaps this commit is the wrong solution though, and the correct solution is to _remove_ the call to `LoadDefaultAssets`.  However it seems likely that Initialize still needs to be called regardless of whether we want to load the default assets or not.

kf6kjg:
We'd had some discussion about a way to flag/detect that whatever loading of default assets it needed had been done - I know I was working on a solution to it at some point.

jimtarber:
I believe the earlier discussion was to move the XML files into a parallel folder after loading (e.g. AssetsInstalled or something like that) as each file was processed, so that they are not found the next time the region starts.

I think that would be better than having dead code that _looks like_ it should work, but doesn't, only because Initialize isn't called.  And also, my concern was that the SetupConnections method wasn't called either.

kf6kjg:
I've just submitted #288 to solve the slow load problem a different way.

jimtarber:
PR #288 does seem to completely avoid the slow LoadDefaultAssets call, so this one is probably ready now as-is.

kf6kjg:
Another option that would enable thus PR is if IW disabled LegacySupport and finally dropped usage of WHIP.

PR #288, opened 2016-10-09
Made a simple change to be able to mark the default assets as loaded.

kf6kjg:
This should dramatically speed up relaunch times.  If you are operating an existing grid and don't want the default assets being loaded, simply create a new file in the assets folder named AssetSets.xml.loaded

Below you can find the list of ideas for solving this annoyance.  This gives some background into why this solution was chosen.
The last idea is the best one, but is a LOT more complex of a solution.  Someday maybe.

== XML->HashUUID->AssetServer
* hash xml file into Hash-style UUID with a special name for possible later cleanup ops and store on asset server.
* If store fails (duplicate) then don't reload the data.

Issues:
* Requires writing a hasher that does the hash-style UUIDs.  Complex.
* Adds MORE ASSETS to the asset server.  Yuck. 

== Move the files to an installed folder
* Really only need to move the Assets.xml file
* Simple move file change.

Issues:
* Git now thinks the file's been removed and wants you to commit the change. (critical problem)
* Requires the server to have write on the folder, which it might not.

== Rename the Assets.xml file
* Simple rename file change.

Issues:
* Git now thinks the file's been removed and wants you to commit the change. (critical problem)
* Requires the server to have write on the folder, which it might not.

== Use a signpost file (current favorite)
* Simple does-file-exist check is only addition to region source.
* Signpost can be added to .gitignore

Issues:
* Requires the server to have write on the folder, which it might not.  Not too big of an issue: the server will just be slow unles the sysadmin manually puts in the signpost file.
* If sysadmin changes any of the default assets he has to 

== Make default asset loading not the region's problem
* Region gets something that doesn't belong to it taken away.
* A special tool would be created for loading in default assets.
* Documentation about the change.

Issues:
* Most complex change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -218,6 +218,7 @@ bin/trustednetworks.txt
 bin/addin-db-001/
 bin/addin-db-002/
 bin/ScriptEngines/Phlox/states/script_states.db3
+bin/assets/AssetSets.xml.loaded
 
 # Other files that need to be created per-installation
 bin/Halcyon.ini
--- a/OpenSim/Framework/Communications/Cache/WHIPAssetClient.cs
+++ b/OpenSim/Framework/Communications/Cache/WHIPAssetClient.cs
@@ -30,6 +30,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Text;
 using OpenSim.Framework;
 using log4net;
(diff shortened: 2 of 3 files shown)


jimtarber:
This looks to be dependent on my other PR #286 
Did you test it with mine?

jimtarber:
When we discussed this previously, we talked of moving the whole folder. The reason for this was for the incremental changes. If only the "root" XML file is renamed, it leaves things awkward if someone wants to add one more asset, like one more texture.  The textures subfile will still exist, but adding the asset won't do anything. Unless the original root file is renamed back to it's original name.

I suppose that if that is done, it is not _too_ wasteful to try to add _all_ of the assets back again, because it will only be done once and then the root file will be renamed as done once again.

kf6kjg:
This wasn't coded with your pr as a dependency.

As to the technique I used, you can see my reasoning above. I pulled up our old conversation and the various ideas discussed then were what I used to create the above list. 

Moving or renaming any file that is under source control results in the possibility of committing that change. So instead I took the route of adding a new file that flags the status and modifying the documentation accordingly.

jimtarber:
Did you test this at all? Without my change, I don't think LoadDefaultAssets is even called, so I don't see how your new code would even be triggered.

jimtarber:
On the technique, if it moved (all) files as completed, we'd need to move the _source-controlled_ location to a different location (perhaps the after-move destination) in order to avoid those issues. On the first run of a new server, to get the assets, that tree could be manually moved to the assets folder and the server run would then put them back in place for source control, once completed.

I was just pointing out there's an alternative, and that moving the single root xml file has some undesirable qualities you didn't mention.  But my concern was with having to add the full set on any change, because using a single file would always load the full set if one asset was added.  But I think that's a weak point since assets are rarely added and it's not a big deal to run the full set when that happens.  So I'm okay with that point, although LoadDefaultAssets is still going to need to be called.

kf6kjg:
In my configuration, MySQL+WHIP only, this code does execute, and speeds up boot up nicely.

Also, in the current technique I'm not moving or renaming anything, I'm creating a new empty file and checking for its existence.

jimtarber:
Ah yes, I spent too much time looking at the notes and not enough looking at the actual code (it's one new file). Looks good, and also +1 for making sure the file create was in an exception handler.

This change won't apply on InWorldz and my test server (where CloudFiles assets are enabled) without #286 also being applied.  However, this PR probably _enables_ that one to be merged, so thanks for the timely PR.

PR #289, opened 2016-10-13
Slow apply physics

jimtarber:
Diagnostics to report slow/inefficient operations in `ApplyPhysics` to attempt to get to the bottom of the high CPU usage on crossings.


kf6kjg:
Looks good to me.  That utility can be useful in many places even if only during dev: I could have used it when chewing on #257.

jimtarber:
I'm going to have some more slow reports to add, and yes it is mostly for experimental dev testing.  However, I the tests also need to include a merged #270.

PR #290, opened 2016-10-15
A new call to llSitTarget must preserve any seated avatar!

jimtarber:
Oops.  A new sit target _offset_ doesn't really mean a new sit target _object_, with _no sitter_.  Fixes report by Trouble Morgan in the forum: http://inworldz.com/forums/viewtopic.php?p=211615#p211615


jimtarber:
This is actually a regression in the 0.9.28 update, if llSitTarget is called _while an avatar is seated_ on the target prim.

PR #296, opened 2016-10-17
Changes to match SL's banned parcel implementation

jimtarber:
As outlined in Mantis 3250:
- minimum height over a parcel _where avatar is in ban list_ now 5000m + terrain height (much higher)
- minimum height over _a parcel with public access disabled_ is now 50m + terrain height (generally lower)

It used to be a fixed 100m for both of those cases.  This also required splitting the constant value into two different values, and adding code to determine which case applied.


ddaeschler:
Nice cleanups on this one too. Looks good

PR #297, opened 2016-10-18
Fixed shallow copy (reference copying) in SOG.Copy

jimtarber:
Several SOG members are references to other objects and the MemberwiseClone() copies the _reference_. So changes to things like sit target lists, or LSL scripted targets, bounding box, etc in one copy of a SOG affected all copies of that SOG that were shift-copied. This resulted in Mike's report of sit targets being cleared in the original _after_ a shift-copy, and is probably responsible for some of the LSL target-related bugs, as well as other object "quantum entanglement" reports.

Reviewed SceneObjectGroup and EntityBase for other possible cases of this shallow copy problem pattern. Didn't find any further cases but did notice that the SOG.Name field (defined in EntityBase but overridden with the RootPart.Name in SOG.Name) wasn't kept in sync on changes. So a second commit also means that changes to the SOG.Name will now also update the object name to keep it in sync with the root part name. This mostly only affects debugging and other symbolic dumps since the Name getter override in SOG pulls the name from the root part.  But this keeps the objects consistent just in case.

### DIFF
--- a/OpenSim/Region/Framework/Scenes/SceneObjectGroup.cs
+++ b/OpenSim/Region/Framework/Scenes/SceneObjectGroup.cs
@@ -209,7 +209,13 @@ namespace OpenSim.Region.Framework.Scenes
                     return String.Empty;
                 return RootPart.Name;
             }
-            set { RootPart.Name = value; }
+            set {
+                RootPart.Name = value;
+                // Also update the object name to keep it in sync with the root part name.
+                // This mostly only affects debugging since the Name getter override above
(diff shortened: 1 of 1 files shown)


jimtarber:
This is an important PR for 0.9.29 as it addresses a regression with sit targets, as well as several other possible problems with LSL scripted targets, bounding box calculations for objects, and corresponding locks which may have been previous shared across multiple shift-copied object instances.

ddaeschler:
Looks good.

PR #299, opened 2016-10-26
Salt those new passwords!

kf6kjg:
Previously new accounts created using the built-in AddUser function left the salt blank, as did the password reset command.
This now actually builds a real salt.  The alphabet could be extended, but that should be sufficient.


jimtarber:
Looks great, nice catch.

PR #309, opened 2016-12-17
Send kill updates if an object moves outside of the view distance.

MatthewBeardmore:
Bug 0002847: Ghost copies of sim-crossing objects
Bug 0002025: DD Culling confusing viewers

Send kill updates if an object moves outside of the view distance.

### DIFF
--- a/InWorldz/InWorldz.Testing/MockClientAPI.cs
+++ b/InWorldz/InWorldz.Testing/MockClientAPI.cs
@@ -614,17 +614,27 @@ namespace InWorldz.Testing
 
         public void SendStartPingCheck(byte seq)
         {
-            
+
         }
 
         public void SendKillObject(ulong regionHandle, uint localID)
         {
-            
+
         }
 
-        public void SendKillObjects(ulong regionHandle, uint[] localID)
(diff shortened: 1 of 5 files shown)


appurist:
Both of these commits look good to me. Thanks Matt!

zauberparacelsus:
@MatthewBeardmore Is change 2506f5b for killing objects when an avatar moves far enough that they leave draw distance, or is it for killing avatars when they leave another avatar's view distance?  I'm afraid I can't tell from the code alone because I'm not familiar with it.

If it is the latter (killing avatars that leave draw distance), it should probably be investigated a bit more.  It could potentially mean that viewer radars would be broken, not showing distant avatars, which would make it more difficult for land owners to moderate their land or deal with abusive avatars.

ddaeschler:
Definitely needs testing for the avatar case, but otherwise looks good. 

I remember we weren't killing objects as the avatar moves because that would require a full culling check and we don't have octrees, but this case where the object moves itself out of draw distance seems safe. The SL simulator sends kills in the same fashion for objects so this should be ok.

Thanks for the commit, and sorry for the delay!

PR #310, opened 2016-12-17
Assorted fixes

MatthewBeardmore:
llClearPrimMedia has no effect with ALL_SIDES
llSetKeyframedMotion does not work on phantom objects
Cannot delete scripts in objects in regions if scripts are disabled on debug page of region window


appurist:
All of these changes look good to me. Thanks again Matt!

PR #312, opened 2016-12-25
Attempted fix for mantis #3274

zauberparacelsus:
This is an attempted fix for iwMakeNotecard that was reported in mantis #3274, where notecards created with one or more unicode characters would have the characters at the end of the notecard truncated.  EG: if the notecard body is "£100, £200", then the written notecard would be "£100, £2".

The cause of the problem was that iwMakeNotecard was using the _character length_ of the notecard body, when it should have been using its _byte length_ (in utf8 encoding).

I have not tested this as there were a number of compile errors, but it should work in theory.  If anyone has a working setup where Halcyon builds without error, please test for me.

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
@@ -14694,7 +14694,7 @@ namespace InWorldz.Phlox.Engine
                     }
                 }
 
-                int textLength = notecardData.Length;
+                int textLength = Encoding.UTF8.GetByteCount(notecardData.ToString());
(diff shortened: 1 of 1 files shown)


ddaeschler:
Can you pull this into a branch and confirm the fix? @appurist

appurist:
I'm planning on testing before and after and single-stepping as soon as I get freed up with other work. (Maps, almost done, a few 0.9.31 release work items. And I'd like to include this in .31 if possible.) @ddaeschler

ddaeschler:
+1

appurist:
I've finally had a chance to check this over and test it. It looks good, confirmed fixed. Ready for merge.

PR #322, opened 2017-01-25
Stub implementation for Get/Set/Reset Animation Override

mdickson:
Added stubbed routines for Get/Set/Reset Animation Override LSL.


appurist:
The stubs look good. Since this is only stubs, merging right away to allow work to continue with the updated Phlox.

PR #326, opened 2017-01-31
Release 0.9.31

appurist:
Everything so far from the branch, should be no real changes except for the AttachmentPt check fix from a few minutes ago.


appurist:
The Phlox binaries in master are newer and need to stay.  But the branch doesn't have the implementation of the post-0.9.31 methods referenced by those binaries.  Not sure how to resolve the conflicts with the Phlox.DLL/PDB file. Presumably we do that at merge time (just don't merge in the ones from the branch, prefer master)?

appurist:
PR #328 replaces this one.

PR #328, opened 2017-02-03
A round-up merge: All updates from the release-0.9.31 branch

appurist:
Same as branch release-0.9.31 except:
- In order to merge with master, and retain updated Phlox binaries in master, this branch reverts "Updated Phlox binaries from the build machine."  (Reverts f8fd555a4f39dd029144bdb1f5bcef304c3aaac6 to avoid conflict.)


appurist:
Some of these are already in master, but because they were cherry-picked, they show as differences.

appurist:
Oh for cryin' out loud... I reverted the wrong revert (wrong commit number) as the first step in the update above. I'm going to need to revert the revert of the revert now.   >.>   <.<

appurist:
The culling fix which caused a regression has been restored so that this branch can include the fix for that for master.  

The "ghost copies" commit is reverted by the `release-0.9.31` branch due to the regression it introduces. But we now have a fix for that regression and don't want to revert it in `master`.

So this special `release-0.9.31-for-merge` branch has everything from `release-0.9.31`, but also with merge conflicts with master resolved, plus it _restores_ this culling change, but now also adds the follow-up _fix_ for that regression. (Phwew.)

This is confusing, but the executive summary is that this fix is cherry-picked into this release so that we can move forward. By reverting the revert, we can merge the fix in here, and get it (in chronological order) into `master`, without conflicts.

appurist:
Summary: This branch is now ready for merge and includes everything from 0.9.31 and also it _does_ include the culling fix, and its followup fix. It does not include any Phlox binary changes (that's what most of the reverts are about) so there are no conflicts now.

PR #329, opened 2017-02-19
Teleport Requests

cinderblocks:
This should allow teleport requests from one agent to another to go through as defined in https://jira.secondlife.com/browse/STORM-1838 and supported by most viewers.

Fulfills http://bugs.inworldz.com/mantis/view.php?id=3282


appurist:
This is safe, and I've tested it.  This change allows "Request Teleport" IM requests from viewer sessions to reach other viewer users.

PR #330, opened 2017-02-19
Fix KillObject message flood preventing some logins (regression in #309)

appurist:


appurist:
This commit is dependent on commit that is reverted by the `release-0.9.31` branch due to the regression it introduces. But we now have a fix for that regression and don't want to revert it in `master`.

So the special `release-0.9.31-for-merge` branch has everything from `release-0.9.31`, but also with merge conflicts with master resolved, plus it _restores_ this culling change, but now also adds the follow-up _fix_ for that regression. (Phwew.)

This is confusing, but the executive summary is that this PR can now be ignored and closed because this fix is cherry-picked into the `release-0.9.31-for-merge branch`.  By reverting the revert, we can merge the fix in there, and get it (in chronological order) into `master`.

PR #338, opened 2017-03-04
Attempt to fix script engine crash seen on SB5, SB6 and Plus SB.

appurist:
- Avoid a race condition where the viewer might send (or a script might request) what results in a second _Replace(itemID, null) for a delete which is a *duplicate*.  This can happen if the user deletes the TaskInventoryItem a *second* time before the viewer updates the Contents to remove it. There must be a script-based equivalent.  This change just handle the itemID no longer being present in the items list. 
See [the forum post](https://inworldz.zendesk.com/hc/en-us/community/posts/115001158408-Sandboxes-updated-to-0-9-32-test-version) reporting it. 
- Also replaced the magic numbers with InventoryType constants.

### DIFF
--- a/OpenSim/Region/Framework/Scenes/SceneObjectPartInventory.cs
+++ b/OpenSim/Region/Framework/Scenes/SceneObjectPartInventory.cs
@@ -700,17 +700,18 @@ namespace OpenSim.Region.Framework.Scenes
             // for a delete which is a *duplicate*. This can happen if the user deletes the TaskInventoryItem 
             // a *second* time before the viewer updates the Contents to remove it. This change just handle 
             // the itemID no longer being present in the items list.
-            if (!m_items.ContainsKey(itemID))
+            TaskInventoryItem item;
+            if (!m_items.TryGetValue(itemID, out item))
                 return;
 
-            InventoryType type = (InventoryType) m_items[itemID].InvType;
+            InventoryType type = (InventoryType) item.InvType;
             if (type == InventoryType.LSL) // Script
                 scriptUpdate = true;
             if ((replaceArgs != null) && (replaceArgs.ReplacementItem != null))
                 newItem = replaceArgs.ReplacementItem;
 
-            m_items.Remove(itemID);
-            _removedItems.Add(itemID);
+            if (m_items.Remove(itemID))
+                _removedItems.Add(itemID);
             m_inventorySerial++;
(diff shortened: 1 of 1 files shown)


OpenSim/Region/Framework/Scenes/SceneObjectPartInventory.cs
```diff
@@ -695,8 +695,16 @@ private void _ReplaceItem(SceneObjectPart part, UUID itemID, bool allowedDrop, b
...
             TaskInventoryItem newItem = null;
             bool scriptUpdate = false;
-            int type = m_items[itemID].InvType;
-            if (type == 10) // Script
+
+            // Avoid a race condition where the viewer might send what results in a second _Replace(itemID, null) 
+            // for a delete which is a *duplicate*. This can happen if the user deletes the TaskInventoryItem 
+            // a *second* time before the viewer updates the Contents to remove it. This change just handle 
+            // the itemID no longer being present in the items list.
+            if (!m_items.ContainsKey(itemID))
```
ddaeschler: This still leaves the potential for the crash to happen on the following InventoryType type = (InventoryType) m_items[itemID].InvType call. I suggest instead using if (m_items.TryGetValue()) and only proceeding if the value exists

ddaeschler:
Race window is shortened, but isn't fixed. See my inline comment

OpenSim/Region/Framework/Scenes/SceneObjectPartInventory.cs
appurist: On further review, I agree. I was concerned that some of the later code, which still uses the itemID, not a reference, might also throw (and another issue or two), so my goal here wasn't as idealist as to eliminate the race condition completely, just to decrease the rate which seemed to increase recently. I also did _not_ want to add any new locks, or change the chronological order of operations here.  However, it seems that the .Remove is safe to call when the itemID is not in the dictionary, and the only real window remaining if we pull the item reference and use that is the .Add (which might add it twice). I'm going to give this one a second kick at the can to see if I can't improve it more without adding risk.

appurist:
I've committed another change that should take care of the rest.  It's still possible to end up in the second half of that function on a duplicate request, but I don't think the effects of that problem warrant more changes, and we may be past the point of pragmatism on this one now.

ddaeschler:
Looks good!

PR #340, opened 2017-03-07
Basic Freeswitch Functionality

mjheilmann:
This code contains updates and modifications to the Freeswitch module present in Halcyon.  The module has been pulled out into a library and refactored to mirror the Vivox module.  It enables those with the inclination or requirement to host their own voice solution to use Freeswitch.

This module enables only basic functionality:  Multiple avatars in a region can connect, and talk/hear using the normal client interface.  Moving between regions switches the avatar to a new channel for that region.  Audio is not directional, there is no active speakers list, and users can only see their own voice dot above their avatar's head.  It is in active use on the MOSES grid for weekly office hours.


appurist:
This is a really great addition to have. Also to have the old Freeswitch code cleaned out.  Looks good to me, @ddaeschler ready for merge (and inclusion in the 0.9.33 release).

PR #341, opened 2017-03-07
JWT mods

mjheilmann:
This pull request includes the following two changes:

1 - Incoming JWT tokens have their signatures verified before they are decoded

2 - The module can now also decode modified base64 strings, which can come from other issuers such as nodejs.

### DIFF
--- a/InWorldz/JWT/JWTSignatureUtil.cs
+++ b/InWorldz/JWT/JWTSignatureUtil.cs
@@ -86,7 +86,7 @@ namespace InWorldz.JWT
         private static byte[] DecodeBase64(string body)
         {
             // Thank you to http://stackoverflow.com/a/9301545
-            body = body.Trim().Replace(" ", "+");
+            body = body.Trim().Replace(" ", "+").Replace('-', '+').Replace('_', '/');
             if (body.Length % 4 > 0)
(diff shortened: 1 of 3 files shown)


appurist:
This looks like an excellent update, some nice catches in here. Thanks!

I do have one question, about an `else` on the `HasValidSignature` check. My concern is that it leaves Payload set to the previous value, it should probably clear it, and `throw new JWTokenException("Invalid Signature");` or something like that?

mjheilmann:
That `HasValidSignature` check occurs within the constructor so it will leave it uninitialized, which is null.

I think you are correct that throwing instead of setting a flag would be better.  I will make that change and examine both RemoteAdmin and RemoteConsole where it is used.

appurist:
Good point. And yeah, not sure what would work best here with the else case, but I noticed that the remote admin change already throws on an invalid signature, so I think that's one reason this one made me think about it. Whatever you think is best there.

mjheilmann:
I made the change, additionally checking the expiration within the constructor.  The now superfluous HasValidSignature and IsNotExpired flags are removed.

appurist:
Very thorough and much simpler! Ready for merge.

PR #342, opened 2017-03-07
Insert additional auth mechanism for remote admin using JWT

mjheilmann:
The RemoteAdmin functionality already has a login_with_password mechanism that uses the Windows Credentials in order to start an xmlrpc session.  I have added an additional login_with_token to allow for JWT token-based authorization against this endpoint.  It operates similarly to the JWT for the RemoteConsole, but has the JWT scope remote-admin instead of remote-console.

### DIFF
--- a/InWorldz/InWorldz.RemoteAdmin/RemoteAdmin.cs
+++ b/InWorldz/InWorldz.RemoteAdmin/RemoteAdmin.cs
@@ -41,6 +41,7 @@ using Nini.Config;
 using Nwc.XmlRpc;
 
 using OpenMetaverse;
+using InWorldz.JWT;
 
 using OpenSim;
 using OpenSim.Framework;
@@ -64,11 +65,14 @@ namespace InWorldz.RemoteAdmin
         private Dictionary<string, Dictionary<string, XmlMethodHandler>> m_commands = 
             new Dictionary<string, Dictionary<string, XmlMethodHandler>>();
 
+        private JWTSignatureUtil m_sigUtil;
+
         public delegate object XmlMethodHandler(IList args, IPEndPoint client);
 
         public RemoteAdmin()
         {
             AddCommand("session", "login_with_password", SessionLoginWithPassword);
+            AddCommand("session", "login_with_token", SessionLoginWithToken);
             AddCommand("session", "logout", SessionLogout);
 
             AddCommand("Console", "Command", ConsoleCommandHandler);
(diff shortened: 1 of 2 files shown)


appurist:
This one looks good too. My only comment would be that maybe the `"./server.crt"` should be configurable, but that's more like a possible future enhancement. This one looks ready for merge to me too.

mjheilmann:
This pull request will no longer compile, as the JWToken class has been modified in another pull request.  Once that is sorted I will update this one.

mjheilmann:
This pull request has been updated to match the changes in the `JWT mods` pull request.  `JWT mods` must be merged as well or token validity and expiration will not be checked correctly.

appurist:
[This line 85](https://github.com/InWorldz/halcyon/pull/342/files#diff-b63a163d2c6d9182889564994e065dffR85) change with the hard-coded `"./server.crt"` reference triggers a System.Security.Cryptography.CryptographicException if the file does not exist. It is not caught, and causes a fatal startup error for any Halcyon installation that doesn't have a server.crt file.

We're going to need a further commit that makes that optional, so we can kill two birds with one stone by making the name configurable and skip this code if none is specified (or an empty string). I'll take care of that right away, unless @mjheilmann is around today and wants to take care of it.

appurist:
There are already Halcyon.ini options defined in the `[Network]` section for the certs used for the HTTP connections; we could/should probably use those unless the JWT certs can be different. If we did use those, from what I can tell from the source code, it uses a different INI option for a cert with a password (`http_listener_ssl_cert`) than one that does not have one (`http_listener_cn`).  Then `http_listener_ssl_passwd` for the actual password.  And it _always_ uses the `http_listener_cn` version if it's defined (ignoring the password), otherwise it passes `http_listener_ssl_cert` and `http_listener_ssl_passwd` (even if no password is defined).

So I think the way I'd like to see this handled is for use to define a new `SSLCertFile`, `SSLPrivateKeyFile` and `SSLPrivateKeyPassword` in the `[Network]` section. Then we can have the code that references the `http_` definitions fall back to those if there are no `http_` definitions don't override them.

appurist:
I've applied this in #346 which covers RemoteAdmin and RemoteConsole (although it does not include the hard-coded names in JWTUserAuthenticationGateway.cs)

appurist:
We've replaced #346 with #348 now. ;)

PR #346, opened 2017-03-19
Configurable (and optional) server.crt file for RemoteAdmin and RemoteAccess

appurist:
RemoteAdmin and RemoteConsole now read the server.crt filename from `[Network]` option `SSLCertFile`
This configuration is optional, then it doesn't initialize a JWT signature util object.


appurist:
In a separate discussion, we decided it was better to pass the filenames in rather than put a direct dependency on Halcyon.ini in the code, so please used #348 rather than this one. (Closing it.)

PR #347, opened 2017-03-19
deadcode

kf6kjg:
Dead code needs to not bother the living.

HttpServerBase, ServicesServerBase, and ProtocolVersions are not used anywhere for any reason.  No zombies.


mdickson:
Actually these are designed for use with other services (like grid services) but the code was forked from original opensim before that work was completed.  As you say I don't believe they are currently used but I'm not sure they're a bad idea either,  The whole Services area deserves a rework so you can eventually split various elements into separate microservices.  But then you, probably, wouldn't do it this way anyway.

PR #348, opened 2017-03-20
Configurable JWT certificate paths and documentation update

kf6kjg:
JWT Certificate path config options, fixups, and docs.

This makes the administrator able to specify the path for the certificates used to verify and create JWTs.  It also updates the docs for the XMLRPC-based remote-admin.


kf6kjg:
Similar to 346, but full spectrum.

Designed to be invisible to servers that do not have JWT certificates.  Those that do will need to edit their XML and region configs to re-enable them.

appurist:
@ddaeschler This PR resolves the current problem in `master` that causes a fatal exception at startup of any region that does not have an SSL cert installed.

PR #349, opened 2017-03-20
Branding scrub

kf6kjg:
Inworldz/opensim cleaned up to more generic or configurable modes.


kf6kjg:
And the typo was cleaned up :)

PR #350, opened 2017-03-22
Inventory folder fetch no longer follows and includes contents of links

appurist:
I noticed that folder contents fetches were actually following and expanding links, both to items and other folders.  I am pretty confident that just retrieving the folder descendants listing should _not_ include the _actual_ items linked to, especially in the case of _folder_ links. 

In both cases, this probably results in both the link _and_ the original being present in the viewer's folder content, which could be particularly troublesome for folders since their parent may not yet even be loaded at the viewer end when we send this subfolder contents down (and inside the folder of another folder other than it's actual parent folder). This change just sends links as links, when filling a folder list, and does not attempt to follow them.

This problem presents itself as unexpected missing inventory item error messages / exception reports trying to follow dead links, even when the viewer is not trying to follow them or do anything with them. 

I've compared the inventory caches from SL and IW and this change seems consistent with those. I've also done some viewer-side testing with the change and found only that the viewer seems very happy with these unfollowed links, correctly using them, or showing them as broken links where the referred-to item has been deleted.  Errors are now only produced when the viewer actually references them, which is protected in most cases but can be visible if the broken link is inside an outfit which is worn.

### DIFF
--- a/OpenSim/Region/CoreModules/Capabilities/InventoryCapsModule.cs
+++ b/OpenSim/Region/CoreModules/Capabilities/InventoryCapsModule.cs
@@ -1045,38 +1045,6 @@ namespace OpenSim.Region.CoreModules.Capabilities
                         {
                             item.SerializeToLLSD(contents);
                             count++; 
-                                
-                            if (item.AssetType == (int)AssetType.LinkFolder)
-                            {
-                                // Add this when we do categories below
-                                linkedFolders.Add(item.AssetID);
-                            }
-                            else if (item.AssetType == (int)AssetType.Link)
-                            {
-                                try 
-                                {
-                                    InventoryItemBase linkedItem = m_checkedStorageProvider.GetItem(m_agentID, item.AssetID, UUID.Zero);
-
-                                    if (linkedItem != null)
-                                    {
-                                        linkedItem.SerializeToLLSD(contents);
-                                    }
-                                    else
-                                    {
-                                        m_log.ErrorFormat(
-                                            "[CAPS/INVENTORY] Failed to resolve link to item {0} for {1}",
-                                            item.AssetID, m_Caps.AgentID);
-                                    }
-                                    // Don't add it to the count. It was accounted for with the link.
-                                    //count++;
-                                }
-                                catch (Exception e)
-                                {
-                                    m_log.ErrorFormat(
-                                        "[CAPS/INVENTORY] Failed to resolve link to item {0} for {1}: {2}",
-                                        item.AssetID, m_Caps.AgentID, e.Message);
-                                }
-                            }
                         } 
                     }
 
@@ -1085,35 +1053,6 @@ namespace OpenSim.Region.CoreModules.Capabilities
                     contents.WriteKey("categories"); //Start array cats
                     contents.WriteStartArray("categories"); //We don't send any folders
 
-                    // If there were linked folders include the folders referenced here
-                    if (linkedFolders.Count > 0)
-                    {
-                        foreach (UUID linkedFolderID in linkedFolders)
-                        {
-                            try
-                            {
-                                InventoryFolderBase linkedFolder = m_checkedStorageProvider.GetFolderAttributes(m_agentID, linkedFolderID);
-                                if (linkedFolder != null)
-                                {
-                                    linkedFolder.SerializeToLLSD(contents);
-                                    // Don't add it to the count.. it was accounted for with the link.
-                                    //count++;
-                                }
-                            }
-                            catch (InventoryObjectMissingException)
-                            {
-                                m_log.ErrorFormat("[CAPS/INVENTORY] Failed to resolve link to folder {0} for {1}",
-                                    linkedFolderID, m_agentID);
-                            }
-                            catch (Exception e)
-                            {
-                                m_log.ErrorFormat(
-                                    "[CAPS/INVENTORY] Failed to resolve link to folder {0} for {1}: {2}",
-                                    linkedFolderID, m_agentID, e);
-                            }
-                        }
-                    }
-                        
                     if (fetchFolders)
                     {
                         foreach (InventorySubFolderBase subFolder in folder.SubFolders)


mdickson:
As I noted in a separate chat discussion, I'm not at all convinced this fixes anything at all related to the current issue being seen on the grid.  As you indicated when we talked "None of that InvCaps code changed since your original commit back in Sept 2012".  When I did the work originally I looked at both server and viewer source code and determined that behavior was required.  Now its true that SL has since moved to a different mechanism for inventory and its possible the viewer code has "rotted" a bit.  But I honestly feel this is a red herring and not at all related to the current issue (changing outfits hangs with the activity cursors spinning, relog required to recover).  

it's just as likely the issue is our past attempts to "clean up" the COF from the server when it should be maintained by the viewer.  It's also possible the issue is troubles in InWorldz around performance of the inventory service.  In any event it seems VERY likely this change doesn't actually fix anything and since there are NO tests accompanying it to validate any sort of fix I don't believe it should be accepted into HEAD.

appurist:
The reason I was even looking at this code was because of the serious problems Calliope was having with inventory, and the log obtained. It was difficult to sort out the real issues because the exceptions captured by the lines deleted in this commit were going off inappropriately (e.g. at login). This makes logs much more difficult to read and consoles completely impossible to do anything.  So I began working through the error reports in Calliope's log, and this one was first. This commit definitely fixes the excessive and unnecessary exceptions of the server following every link item and folder during an folder contents _list_ inventory fetch.  It's going to need to do that if something is rezzed, or loaded, but that is a different operation and it is not needed to fetch the link as an item in the list.  So at the very least, this commit fixes the excessive reporting of errors that has troubled us since the move to Cassandra resulted in these cases throwing exceptions.

It may or may not help with viewer-side problems. I don't know yet, it is why I asked Astoria last night if she would be able to test a new build on the beta grid; if she could reproduce [her inventory issues](https://bugs.inworldz.com/mantis/view.php?id=3162) there. We may be able to get more definitive answers on that issue, or it may have no effect at all, but that is part of the testing of this change.

Regarding testing, where are the current viewer-side (e2e) tests; I'd be happy to update them. You're worried about the effect on the viewer here so how exactly do you propose this be tested other than the way I have been?  There is no lack of testing here; I spent most of the evening testing different combinations, with one or two avatars present, with relogs, outfit changes, cache flushes, etc.  I also checked the UUID.inv.gz cache from SL and compared it to the InWorldz result; as far as I can tell they are identical when it comes to links.  There is one area of testing which I have not done enough of yet, and that is testing the old code to confirm whether the viewer safely discards the extra data, or finds it as a duplicate based on item ID, or gets very confused if the inventory tree arrives out of parentage order, etc.

There is a huge potential for major viewer-side inventory problems here, by including following the folder links. In the COF, folder links are very significant and suggesting that we just ignore this find because you are not convinced that it fixes anything is not a productive suggestion.

Speculation that there are other problems as well is something I can agree with, but it doesn't mean we shouldn't follow up on each issue as we find them. I'm looking through the log of problems Calliope has with the current unchanged code, and that I also see with my own avatar on logins to my test server.  It's easily reproducible before this commit, and is not, after the commit.

I think further testing with Astoria would be good; it may show that her problems remain, or it may show that they decrease or go away.  But I can see the kinds of problems this extra data could cause at the viewer end.  I will do more A/B testing and debugging today, in particular, testing the effect that the existing (pre-commit) code has on viewers with behaviors and debugging the viewer handler.

appurist:
In addition to the actual functional reliability tests, I've done some additional A/B testing comparing the current release server vs the server after this PR is applied.

I've removed the `UUID.inv.gz` files and logged in with the IW 3.0.2 viewer, logged out, and then saved the resulting `UUID.inv.gz` inventory cache file.  I then repeated the process with server code that has this PR applied.

I've examined the resulting `UUID.inv.gz` files, and diffed them as well, and the results are a bit surprising:

- The most notable difference is that the older unpatched code results in an inventory cache that does not include the `Current Outfit` folder, at all.  See [this image](https://i.gyazo.com/5aea88873e11dec1697026070ef9ef6e.png) from a diff.  I thought maybe it had just been moved or stored in the wrong spot in the folder tree, but that ID (`b2fcc515-9423-5ceb-5f78-1ddac3ce3497`) does not appear in the inventory cache under the old server at all.

- Some of the other differences are inconsequential, e.g. some of the gesture items are in a slightly different order.  I'm thinking this is just a sign that part of the process here is async, and small order differences may occur.

- The two inventory cache files have different sizes, more than just the size represented by the `Current Outfit` folder.  Old: 3,086,565 bytes, new: 3,096,030 bytes. That's a difference of 9,465 bytes. After a casual examination of the differences, they appear to be mostly (or only) wearables that are missing from the old inventory cache.

I did not notice any observable differences from the perspective of my viewer session. Both logins appeared to operate normally, but clearly the viewer data was incomplete under the current code.  I suspect this may be related to including actual folders whose parent has not yet been delivered via the inventory fetch, thus orphaning the folder and resulting in it not being saved at exit, and perhaps not used correctly during the viewer session.

I am not sure what other problems might be result from doing an in-place macro expansion of the links, but to me this is enough to confirm my theory that the viewer is not able to properly interpret these inventory lists.

And since item and folder links are common for the `Current Outfit` folder, and it and wearables are the most notable absentee from the cache, I would think there is a very good chance that this may be related to at least some of the issues reported with outfits and inventory.

appurist:
Actually I now understand and can explain why, in the current released server, the viewer cache is missing the Current Outfit folder (COF) and perhaps others:

Since the current released code sends both the link and the item/folder it is pointing to, the viewer may receive _too many_ items for any folder (such as COF) that has at least one link in it.  And the viewer's `isCategoryComplete` function checks not for `actual >= expected`, but `actual == expected`. So if it receives too many, it thinks the inventory fill is still incomplete.

See this viewer function, and the `==` on line 1887:
isCategoryComplete

appurist:
Note: From my perspective, this _confirms_ that this PR is an important fix, and that the viewer is not expecting the extra data, and it is in fact causing the COF and other items to not to be cached on exit. And very likely other negative symptoms during viewer execution.

appurist:
`2017-03-23T01:06:09Z WARNING: LLAppearanceMgr::updateAppearanceFromCOF: COF info is not complete. Version 7718 descendent_count 18 viewer desc count 19`
Viewer testing confirms my theory above: it is receiving too much data. An extra item here causing the COF to be considered incomplete.

PR #355, opened 2017-03-27
Replace prebuild with .sln file and nuget-based libs for some dependencies

appurist:
- Nuget dependencies now include the LibreMetaverse libraries, C5, log4net, nunit, protobuf-net, XmlRpcCS, and zlib.net.
- The nuget dependencies tend to be a newer version of the library. Usually the latest, except for nunit which is still 2.x (although slightly newer now at 2.6.4).  Note: ANTLR3 on nuget is not compatible, so it is nuget is not used for that one. Causes many tests to fail if the nuget version is applied.
- All of the prebuild stuff has been removed. Use the Halcyon.sln now directly.
- Halcyon.exe.config is now auto-created from the app.config on each build and not overwritten/lost.


appurist:
@ddaeschler #354 should get merged before this one, since it modifies the `prebuild.xml` file, and this PR removes that file.  If there are any conflicts just let me know and I'll update this one for merge.

ddaeschler:
Deletion conflict as predicted

appurist:
Conflicts resolved, but please do NOT merge this until I can confirm a login problem seen in beta grid testing isn't part of this merge. (So far, this merge appears to break something in the User login server.)

PR #367, opened 2017-04-21
Fixed presence error caused by MySQL exception on LogoutUsers() at region shutdown

appurist:
Fixed `MySqlException: Data truncated for column 'logoutTime' at row 1` which caused `MySQLUserData.LogoutUsers()` to fail to update the `agents` table on a region shutdown.  The problem was basically trying to stuff a `DateTime` into an `int(11)` SQL field, which caused an SQL overflow exception.


appurist:
Actually it probably didn't affect the presence as simply as I described, but the `agents` table would probably have the wrong region handle. Which was one of the other null reference crashes fixed earlier, in #366 . At any rate, it's an obvious bug that's easy to fix and has been in the Halcyon code since before the move to GitHub.

PR #379, opened 2017-05-04
Release v0.9.34.6334

appurist:
Fixed two regressions in avatar position and a possible one in anim packs:
- avatars not sent on login until avatar movement or camera changes (`InitialDataSent` needed to be set _before_ sent, so now it's `InitialDataReady`)
- child agent update `NO_POSITION` test got inverted in the "nothing" change. Brain fart, got the logic in the `==` and `!=` backwards in the conversion to `CompareTo`. Scrap that, it's harder to read anyway.
- `SendAnimPack` now needs the same agent in region flags as sending the initial data. Avoids allowing anim pack to be sent before the initial avatar data.  Restored the old code there.


appurist:
My testing revealed a further problem, plus I accidentally left a debug message in that spams neighbor regions. It's going to need another commit or two. Closing this PR for now.

PR #380, opened 2017-05-04
Better xml env var substitution

kf6kjg:
More usable feature, plus some minor cleanup.


appurist:
This looks good to me. I really like the solution of using the environment variable if present, but using it as a literal if it doesn't match an environment variable. Especially with the length check.

Looks good, ready for merge.

PR #385, opened 2017-05-13
Culling fix for missing prims after leaving draw distance and returning.

appurist:
This should fix [Mantis 3298](https://bugs.inworldz.com/mantis/view.php?id=3298) and other cases like it.

### DIFF
--- a/OpenSim/Region/CoreModules/Agent/SceneView/SceneView.cs
+++ b/OpenSim/Region/CoreModules/Agent/SceneView/SceneView.cs
@@ -788,12 +788,26 @@ namespace OpenSim.Region.CoreModules.Agent.SceneView
                     {
                         if (m_updateTimes.ContainsKey(part.LocalId))
                         {
-                            m_updateTimes.Remove(part.LocalId);
                             sendKill = true;
+                            // If we are going to send a kill, it is for the complete object.
+                            // We are telling the viewer to nuke everything it knows about ALL of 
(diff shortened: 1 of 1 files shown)


OpenSim/Region/CoreModules/Agent/SceneView/SceneView.cs
```diff
@@ -788,12 +788,26 @@ public void SendPrimUpdates()
...
-                            m_updateTimes.Remove(part.LocalId);
                             sendKill = true;
+                            // If we are going to send a kill, it is for the complete object.
+                            // We are telling the viewer to nuke everything it knows about ALL of 
+                            // the prims, not just the child prim. So we need to remove ALL of the 
+                            // prims from m_updateTimes before continuing.
+                            IReadOnlyCollection<SceneObjectPart> sogPrims = part.ParentGroup.GetParts();
+                            foreach (SceneObjectPart prim in sogPrims)
+                            {
+                                if (m_updateTimes.ContainsKey(prim.LocalId))
```
zwagoth: According to https://msdn.microsoft.com/en-us/library/kabs04ac(v=vs.110).aspx the ContainsKey check is likely an anti-pattern. Remove throws no exception when the key isn't found, so you're duplicating the key exists cost by doing both.

ddaeschler: Good catch! I'll accept a simple PR with this change if you can get one up

PR #386, opened 2017-05-14
PID Files

kf6kjg:
This makes all servers generate PID files just like the User server already does.

It also adds a new commandline flag, `--pidfile=path`, so that external systems can specify where the PID file is to be placed.

Lastly it cleans up the `main()` functions of all three grid services so that they match each other.  This should help with future maintenance, assuming people read the comment I left...


appurist:
This one looks good for merge, but it's difficult to review exhaustively. Has it been tested on Windows and Linux?

PR #387, opened 2017-05-14
Forward slash is the only valid dir separator

kf6kjg:
It even works right in Windows, but the backslash doesn't work right in Linux.


appurist:
This commit is good in that it allows it to run correctly on Linux now (too), but ideally if we're cleaning up cases like this, we should be using `Path.DirectorySeparatorChar` from `System.IO` rather than hard-coding path separators.  Or using `Path.Combine` to join the parts.

PR #388, opened 2017-05-14
Ci integration

kf6kjg:
How would you like your compilation served?  This iteration is just Windows .NET AppVeyor builds and works with runprebuild.bat if it exists, and skips it if it does not.

The version is pulled from the resulting OpenSim.Framework.Servers.dll, so should reflect real-world versioning of Halcyon.

It is set to build on pushes to every branch.  This can be restricted to any specific branch or set thereof.

I can also set it up to push the result to a draft GitHub release if it is a tagged commit.

See example build here: https://ci.appveyor.com/project/kf6kjg/halcyon
And see example resulting artifact ZIP at: https://ci.appveyor.com/project/kf6kjg/halcyon/build/artifacts

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-version: {build}
+version: 0.9.00.{build}
 
 image: Visual Studio 2017
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -20,7 +20,7 @@ after_build:
     - ps: >-
       If(Test-Path "bin/OpenSim.Framework.Servers.dll") {
         $halcyon_version = (ls bin/OpenSim.Framework.Servers.dll | % versioninfo).ProductVersion
-        Update-AppveyorBuild -Version "0.$($halcyon_version.Substring(0, $halcyon_version.lastIndexOf('.'))"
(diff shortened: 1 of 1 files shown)


appurist:
I _love_ seeing progress like this, on the goals that we all know we want to make progress on. CI and CD are clear goals here.

I also like this commit because it is confined to a single file, that has no impact on anything else in the project.

One concern I have is how out of date it could get very quickly with other commits.  We had a full nuget-based build in a [branch](https://github.com/InWorldz/halcyon/commits/nuget-not-prebuild) but we had to postpone that merge due to a couple of issues that popped up with limited time.  Once the missing prim fix goes out in 0.9.35, I think we need to return to the nuget work before the branches get too far out of sync and we loose that work.  At that time, the .sln file will be the "hot" original, and runprebuild will be gone.  I think this PR is nicely set up for that being optional so we're good, but any commits related to the build process will need to be aware of this file and the CI process.  We can discuss the setup for that separately.

I think this PR is good to go, at least in terms of it not having impact on anything done elsewhere, but may need additional commits in the future as we update the build process (or may not!).

kf6kjg:
Actually I designed it to seamlessly handle that branch. :)  Didn't test, so I'm not 100% confident, but it should.

appurist:
"it is confined to a single file, that has no impact on anything else in the project."

Merging. ;)

PR #391, opened 2017-06-20
Incoming HTTP fixes: x-query-string compatibility, null reference exceptions

appurist:
This update fixes problems with:
- null reference exceptions in several places in the code if an incoming HTTP request includes a query string that does not have at least one `=` in it. This fixes `500` error responses on many HTTP requests.
- `llGetHTTPHeader(id, "x-query-string")` returned an empty string if there was a query with no `key=value`, such as `URL/?abc+def+ghi`.
- added `llGetHTTPHeader(id, "x-query-string-compat")` in case anyone still needs the old broken behavior

### DIFF
--- a/OpenSim/Region/CoreModules/Scripting/LSLHttp/UrlModule.cs
+++ b/OpenSim/Region/CoreModules/Scripting/LSLHttp/UrlModule.cs
@@ -506,7 +506,7 @@ namespace OpenSim.Region.CoreModules.Scripting.LSLHttp
                         String[] keys = (String[])de.Value;
                         foreach (String key in keys)
                         {
-                            if (request.ContainsKey(key))
+                            if ((key != null) && request.ContainsKey(key))
                             {
                                 string val = (String)request[key];
                                 queryString = queryString + key + "=" + val + "&";
@@ -518,11 +518,15 @@ namespace OpenSim.Region.CoreModules.Scripting.LSLHttp
                     }
                 }
 
+                // Grab the raw unprocessed original query string, if any.
+                int rawQueryPos = httpRequest.Url.Query.IndexOf('?');
+                string rawQueryStr = (rawQueryPos < 0) ? httpRequest.Url.Query : httpRequest.Url.Query.Substring(rawQueryPos + 1);
+
                 //if this machine is behind DNAT/port forwarding, currently this is being
                 //set to address of port forwarding router
                 requestData.headers["x-remote-ip"] = httpRequest.RemoteIPEndPoint.ToString();
                 requestData.headers["x-path-info"] = pathInfo;
-                requestData.headers["x-query-string"] = requestData.uri;    // raw (SL-compatible)
+                requestData.headers["x-query-string"] = rawQueryStr;        // raw original (SL-compatible)
                 requestData.headers["x-query-string-compat"] = queryString; // processed (old Halcyon scripts)
                 requestData.headers["x-script-url"] = urlData.url;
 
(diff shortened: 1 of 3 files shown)


OpenSim/Region/CoreModules/Scripting/LSLHttp/UrlModule.cs
```diff
@@ -518,11 +518,16 @@ public void AsyncHttpRequest(IHttpServer server, string path, OSHttpRequest http
...
+                int rawQueryPos = httpRequest.Url.Query.IndexOf('?');
+                string rawQueryStr = (rawQueryPos < 0) ? httpRequest.Url.Query : httpRequest.Url.Query.Substring(rawQueryPos + 1);
+
                 //if this machine is behind DNAT/port forwarding, currently this is being
                 //set to address of port forwarding router
                 requestData.headers["x-remote-ip"] = httpRequest.RemoteIPEndPoint.ToString();
                 requestData.headers["x-path-info"] = pathInfo;
-                requestData.headers["x-query-string"] = queryString;
+                requestData.headers["x-query-string"] = rawQueryStr;        // raw original (SL-compatible)
+                requestData.headers["x-query-string-compat"] = queryString; // processed (old Halcyon scripts)
```
kf6kjg: If they are old then they'll not be calling the compat header and will break anyway.  The only reason I can thing to introduce this new header is to provide the old method for new scripts being written to use the old method - not sure there's any need for that, especially since the semantics have changed: the processing now skips nulls.

appurist: That's exactly the reason for adding it.  I'm not sure it will help but it won't hurt to have a compatible code option available should someone need it.

We don't really have any other option here for the normal x-query-string implementation. It's just plain broken in many ways and needs to be fixed, and the fixes produce different results than the old code. So I've added a second option to return the former value. That would require a script update for anything that _relied_ on the old results, although that was a null string anyway, for many of the broken cases.  But anything that expected only an `abc` entry for `abc=def&ghi` at least has an easy method to get what the previous server would have returned.  Whether that will be used or not, I don't know. Just trying to cooperate with the old script logic as much as possible, just in case.

mdickson: I dont see alot of merit in the compat string.  You wouldn't have coded to fetch it by that name and if you needed to recode to handle it you might as well fix your script to use the raw value anyway.

appurist: I think both of you are thinking the same thing and I definitely agree it may never be used. However, the case I'm worried about is one where someone has handed out a free script that is possibly widely used, by people who are **not** scripters, and can't as Mike said "just fix your script".

Given that this is a possibly breaking change, depending on how the existing query string is used, I just want a backup plan in place that allows us to say to non-scripters "Oh your script doesn't work anymore? Okay just change 'x-query-string' to 'x-query-string-compat' and it's back to it's old behavior."

Without the compat alternative, we're opening ourselves to much more serious complaints after an upgrade, that people who aren't necessarily the authors of scripts are "suddenly forced to _rewrite them_ because we broke the server".  Including a backwards-compatible variant avoids all that.  Given it's dubious likelihood of being used, I wouldn't have included it if it had a high cost, but it's effectively free since it's just returning what the old one did.

Yes, it may not be used, but the InWorldz philosophy is always to do everything possible to avoid breaking content, and if that is required in some fix, provide the safest and least disruptive alternative path.  So in this case, better safe than sorry and provide it as a simple text constant change to fix anything that breaks.

appurist: Shorter answer: I don't have any reason at all to avoid providing a backwards-compatible option. Thinking nobody will use it isn't a good reason not to offer it. It's one well-tested line of code (what we always did). Is there any actual _problem_ with providing it in case it's needed? I don't see one.

For code review, I think the changes to the existing query string option is the significant case. It's an incompatible change.  However, it's currently triggering exceptions and 500 errors and producing the wrong results when compared to SL.  I only know about it because we have a specific case where a properly written script is failing in IW due to the bugs in the former implementation.  So I don't see any option there except to make it work, which also means make it different.  But we can at least provide the old string as well, minus the exceptions and 500s.

PR #392, opened 2017-06-25
Report details of RezObject failures to help track down lost CF asset report.

appurist:
Investigating a report of MLP ~ball object that won't rez. Appears to be missing asset. Need the ID.

### DIFF
--- a/OpenSim/Region/Framework/Scenes/Scene.Inventory.cs
+++ b/OpenSim/Region/Framework/Scenes/Scene.Inventory.cs
@@ -4247,7 +4247,14 @@ namespace OpenSim.Region.Framework.Scenes
                         reason = "success";
                         return allGroups;
                     }
+                } else
+                {
+                    m_log.ErrorFormat("[SCENE]: RezObject could not fetch asset {0} for '{1}' in '{2}' owned by {3}", item.AssetID, item.Name, sourcePart.ParentGroup.Name, sourcePart.OwnerID);
                 }
+
+            } else
+            {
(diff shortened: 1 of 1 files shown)


OpenSim/Region/Framework/Scenes/Scene.Inventory.cs
```diff
@@ -4247,7 +4247,14 @@ private bool CheckGreyGoo(SceneObjectPart sourcePart, SceneObjectGroup newGroup)
                         reason = "success";
                         return allGroups;
                     }
+                } else
```
kf6kjg: (nitpick) Isn't it normal in this codebase to drop the else onto it's own line or am I missing something?

appurist: I dunno. I think in general (maybe since some devs have been using Resharper) it seems to be the norm to include the { }. I have skipped it for single-line statements in the past, but more JavaScript lately, tend to wrap simple statements in `if` or `else` with the compound { } markers.

Oh I think you are specifically referring to just a missing newline there.  Another good question. I'm a firm believer of not wasting vertical space, you can understand code better when you can see more of it. So I've actually used `}  else {` in some cases. Not sure why I didn't there.

I don't have a strong preference either way, as long it's fairly consistent. It wasn't a conscious decision.  So I'm not going to do another commit but I also wouldn't object to someone changing the whole file in the future (in a separate commit from whatever other changes they might have wanted to make, otherwise it's impossible to diff the 'real' changes.

To be honest, if I was going to "fix" this, I'd probably drop the newline after the else too, to bring it tighter vertically. ;)

PR #393, opened 2017-07-18
Lowered most notecard delays, new delays on some errors. (Mantis 3304)

appurist:
As per the discussion in [Mantis 3304](https://bugs.inworldz.com/mantis/view.php?id=3304):
- GetNumberOfNotecardLines now includes the old 100ms delay on "notecard not found" error cases.  (Previously no delay at all, even if errors.)
- If cached, GetNumberOfNotecardLines only delays 25ms now instead of 100.
- If asset fetch required (uncached), GetNumberOfNotecardLines only delays 50ms now instead of 100.
- Notecard line/segment fetches now delay only 1ms (probably 16ms) every 16 lines, starting at the first line, if cached.  (Should result in be 1000 lines per second now.)
- Notecard line/segment fetches continue to delay by 25ms if not cached (i.e. first line only), same as above.

### DIFF
--- a/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
+++ b/InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
@@ -14500,6 +14500,9 @@ namespace InWorldz.Phlox.Engine
 
         public string GetNumberOfNotecardLines(SceneObjectPart part, string name)
         {
+            const int ERROR_DELAY = 100;
+            const int LONG_DELAY = 50;
+            const int FAST_DELAY = 25;
             TaskInventoryDictionary itemsDictionary = (TaskInventoryDictionary)part.TaskInventory.Clone();
 
             UUID assetID = UUID.Zero;
@@ -14520,7 +14523,7 @@ namespace InWorldz.Phlox.Engine
             {
                 // => complain loudly, as specified by the LSL docs
                 ScriptShoutError("Notecard '" + name + "' could not be found.");
-
+                ScriptSleep(ERROR_DELAY);
                 return UUID.Zero.ToString();
             }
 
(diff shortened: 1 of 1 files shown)


InWorldz/InWorldz.Phlox.Engine/LSLSystemAPI.cs
```diff
@@ -14612,7 +14618,8 @@ private string GetNotecardSegment(SceneObjectPart part, string name, int line, i
             {
                 AsyncCommands.
                 DataserverPlugin.DataserverReply(reqIdentifier, NotecardCache.GetLine(assetID, line, startOffset, maxLength));
-                ScriptSleep(DELAY);
+                if (((line % LINES_PER_DELAY) == 0) && (startOffset == 0))
```
kf6kjg: While clever and stateless, this runs afoul of an attack vector that trivially defeats it: simply have your script only work with lines that have no delay.  _Et voila_ the whole purpose of this has been nullified.

appurist: That isn't the purpose, and wasn't even one of the goals.  They were:
- reduce the time it takes to read a notecard
- do not completely eliminate the penalty because we occasionally have a buggy script that would then spin in a tight loop

appurist:
I wasn't worried about any "bypass" and in fact the primary goal is _to lower the notecard reading times_ enough that that nobody would ever be encouraged to bypass the load time anyway.  Load times are currently incredibly painful, and anyone who wanted to intentionally bypass it could use many other mechanisms, such as just using many scripts to read lines in parallel, etc.  There are many attack vectors if one chose to be malicious but we haven't seen a single case of that even _with_ the punishing times.  Reducing the load times to 1000 lines per second as above should eliminate any desire to try to circumvent the delays. In fact, they probably would think we've eliminated them.

We discussed eliminating the delays completely, if the notecard was in cache, but I rejected that idea because I was concerned about eliminating a loading penalty from buggy scripts.  It doesn't mitigate all buggy script cases though, so if you have a more complex alternative implementation in mind, feed free to implement one. I tried to come up with something that would not require script state to be modified, because it wasn't a priority and the only way this was going to get improved was if it was low-hanging fruit. If you have time for something more, then by all means... ;)

PR #394, opened 2017-07-20
Fixed exception saving scripts under updated viewers (e.g. Alchemy 5)

appurist:
Newer viewers are using the sim CAP now, which passes a bool, but LLSD doesn't actually use a real bool, it's an int under the covers, so we need to convert or we get a C# type exception. 

This matches the fix in OpenSim for the same lack of support in the server code. This should actually be fixed at the OSD level but until that happens, this change should avoid the exceptions for all LLSD messages that include bools. Thanks to Cinder for this one.


appurist:
I'm going to merge this one. It's not my change, it's the one already applied to OpenSim for the same issue. I've reviewed it and tested it and it fixes a showstopper problem in the current `master` (unable to edit scripts) affecting newer viewers, including the Alchemy 5 release I was testing when I found this. We need a `master` that doesn't crash, and this fix looks good.

PR #396, opened 2017-07-27
Missing attachments on viewer relogs

appurist:
Fixed a possible race condition with session global `_rezMultupleAttachmentsData`.  It is used outside the lock; it might be getting cleared during async processing by another packet, especially if the asset cache isn't warmed up, such as on a different region from where the attachments were added.  Also some small tweaks to some of the inventory/asset reporting to help us investigate issues like this.


appurist:
Initial lengthy testing by @zauberparacelsus and I today were unable to reproduce the issue after testing on a region with these changes. However, after that we weren't able to reproduce the problem even on non-updated regions. So it's a bit unclear if this resolves the case in question, but it certainly didn't make it worse. Going to merge this for the RC deployment to Sandboxes tonight.

PR #398, opened 2017-07-28
Added a user-configurable limit (1 or 2) on the range of neighbor regions.

appurist:
Added a `LimitNeighbors` option to the `PresenceFlags` for region changes, and `/!neighbors range 1` command. Does not affect the range that region servers see, only the range viewers are informed about on region entrance.

The purpose of this is to test whether it would be beneficial to restore the old range of 1 region for neighbors, for bandwidth and connection reasons.  The current value of 2 can result in as many as 25 last regions connected to a viewer session, 30 regions on a crossing, and as many as 50 regions connected (on a teleport).  If a fair number of these have high content in the region, it has crashed viewers and seriously impacted performance. This is in some ways a stop-gap solution until Halcyon servers support view frustums and interest lists, but even then this option may be useful to limit wasted bandwidth when you don't care about neighbors, to limit neighbors during photography or recording, to test various features, to force a reconnection to distant regions, or even to provide region-wide privacy in some cases.

Also removed unused `GetKnownNeighborCountWithinClientDD` function.

### DIFF
--- a/InWorldz/InWorldz.Testing/MockClientAPI.cs
+++ b/InWorldz/InWorldz.Testing/MockClientAPI.cs
@@ -132,6 +132,12 @@ namespace InWorldz.Testing
             set { }
         }
 
+        public uint NeighborsRange
+        {
+            get { return 1; }
+            set { }
+        }
+
         public uint CircuitCode { get; set; }
 
 #pragma warning disable 0067 // disable "X is never used"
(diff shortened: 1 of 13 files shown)


appurist:
I have finished testing this, and I have 90 minutes to build the 0.9.36 RC, so I'm going to merge this for the 0.9.36 build, but feel free to send me any review comments privately or on the `master` commit itself.

PR #399, opened 2017-08-07
Bandwidth usage fixes

appurist:
Fixed bandwidth calculation problems, especially neighbor regions. …
- Fixed the DripRate to throttle value conversion, which confused bits with bytes, causing rates to be 1/8 what they should have been.
- The above problem was because the units of measurement (bps, Kbps, bytes, KBps) were confused throughout most of this code, so I've added a lot more comments, small tweaks to the code to make it more readable, understandable and maintainable. Hopefully the units of measurement are now clear.
- Fixed a total stall of outbound queues if latency was high (which is somewhat intentional on neighbor regions with the 30ms timer).
- Bandwidth is _seriously_ underused, with a cap of 250KB/s at the viewer end, which results from the effective viewer cap of 2000 kbps, which comes from the hard limit of 3000 kbps.  (Note, IW3 allows a larger number in the Preferences form, but still caps it at 3000 when sending to the server.)  So let's allow more for bandwidth to neighbor regions (doubled).  When moving and adding more regions, the current region is probably mostly filled anyway if not completely filled, or will be by the time the avatar hits the border. And when flying or sailing the waterways, neighbor limits are _overall_ limits.  This might change back to 30% with more viewer tuning and higher bandwidth settings (coming soon, Drake is investigating after I reported these issues.) Also became a problem with the neighbors range of 2 increase a few years ago, since that increased neighbors from 8 to 24, and then we divide the neighbors limit by the number of neighbors. 1/24th of 30% is far too slow of a trickle, especially when we use 2/3rds of the bandwidth of other grids, and we also use it to limit textures via Aperture (nobody else does).
- Given the above, I'm also testing a minimum of 10% bandwidth for each neighbor.  The current viewer limits mean even a 5Mbit connection is underused, and most users probably have something better than that. Those with tight Internet connections can use reduced levels (like 1000) and still get some data from all regions.

### DIFF
--- a/OpenSim/Region/ClientStack/LindenUDP/LLClientView.cs
+++ b/OpenSim/Region/ClientStack/LindenUDP/LLClientView.cs
@@ -1077,6 +1077,7 @@ namespace OpenSim.Region.ClientStack.LindenUDP
         private Int32 m_LastQueueFill = 0;
         private Int32 m_maxUpdates = 0;
         private bool m_lastSentFull = false;
+        private Int32 m_prevOutboundQueueSize = 0;
 
         private void EmptyQueueProcessUpdates(ThrottleOutPacketTypeFlags categories)
         {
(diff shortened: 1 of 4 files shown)


appurist:
Merging, this is on the main grid now. It doesn't solve everything but it provides some options and seems to be a general incremental improvement.

PR #404, opened 2017-10-21
Several improvements to asset store operations

appurist:
- `StoreAsset` calls did not use the asset operations threadpool; they were synchronous. `StoreAsset` now precaches the `wireAsset` and immediately returns to the caller once the asset is cached.  Later, asynchronously, it may update the cache with the stream asset, or on error save it to the local cache (as in the past, just async now).
- Also fixed the accounting of sizes in `LRUCache` if an item is being replaced.
- Fixed the excessive "asset not found" reports. `AssetNotFound` is also called from the WHIP client, so all new (CF) assets would produce this report when the operation was in fact a success.


appurist:
I've done all the testing and review I can do on this, merging for beta/Sandbox/other test regions.

PR #406, opened 2017-11-25
Estate Powers for llPushObject

zauberparacelsus:
Changes llPushObject so that it will function on push-restricted land if the script is owned by the estate owner or an estate manager.


appurist:
This PR looks completely fine to me. The only changes are to allow the EO or EM to also use any push that the parcel land owner could do, i.e. to treat EO/EM as owner-equivalent without testing the actual logic. This should be safe and acceptable to all. Merging.

PR #413, opened 2017-12-05
Allow StoreAsset to be optionally async, default disabled (opt-in).

appurist:
Due to subsequent fetches coming only from Aperture (which is unaware of the new asset until the StoreAsset completes the write to CF), we can't signal the session to continue by returning early (as soon as the asset is in the region server cache).

This change effectively reverts the change to make StoreAsset asynchronous, although it leaves it optional in the Halcyon.ini file (default off).

By default, this restores the previous behavior of synchronous CF writes that block the session.

### DIFF
--- a/InWorldz/InWorldz.Data.Assets.Stratus/CloudFilesAssetClient.cs
+++ b/InWorldz/InWorldz.Data.Assets.Stratus/CloudFilesAssetClient.cs
@@ -625,8 +625,15 @@ namespace InWorldz.Data.Assets.Stratus
             statTotal++;
             statPut++;
 
-            // Now do the actual CF storage asychronously so as to not block the caller.
-            _threadPool.QueueWorkItem(() => StoreCFAsset(asset, wireAsset));
+            if (Config.Settings.Instance.UseAsyncStore)
+            {
+                // Now do the actual CF storage asychronously so as to not block the caller.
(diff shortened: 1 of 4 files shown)


appurist:
This is not an option for anyone to use, ever, in the future, except for me during this WIP testing phase. It exists to avoid having to update code on a production grid only to enable testing of this code that already exists in the region server. Other than my specific tests, this will always be hard-coded off, or if we find some way to work around the side effects, it will be hard-coded on and the option will be removed.  However, for now I'll add it to the sample file with a comment to never use it.

mdickson:
If the INI option is developer only is there a good reason to actually include it?  Why not set the change in code and test and decide how to run it before its merged to HEAD.

appurist:
> It exists to avoid having to update code on a production grid only to enable testing of this code that already exists in the region server. 

This is a small no-brainer change that is part of enabling more real-world testing on developer test regions (as we have done many times) such as Zauber's and Mike's regions (and in most cases Sandboxes). 

I'm trying to package up an update that can be tested and will provide all the other commits to the full grid while not eliminating this WIP from the same build.  It will provide us with a disabled version for the Sandboxes and possible rollout while allowing parallel progress on developer test regions.

I'm not looking for feedback on whether this _should_ be done, but rather whether there are any bugs/typos etc. (Although I have verified the change with breakpoints etc before committing.)

PR #421, opened 2018-02-18
Mitigate against a late CompleteMovement retry which is a Bad Thing

appurist:
This may resolve a large number of "login-status-is-completely-screwed" cases, as well as [Mantis #3287](https://bugs.inworldz.com/mantis/view.php?id=3287).  This `CompleteMovement` request is handled with `PacketHandlingDisposition.HandleSynchronized` however it has an async section at the end, and no protection from queuing up the duplicate request for serial processing once the first call is complete. We're tracking `AgentInRegion` status here so just avoid resetting it at the start, and instead, use it to recognize the duplicate.


appurist:
I've done the practical run-time tests as well now, with breakpoints, delays etc. and confirmed it is coming in as None and should not be anything but, unless it's a (previously destructive) redundant duplicate call already processed (or still being finished in the async part).

Thanks for the quick review.

PR #427, opened 2018-03-10
Added additional handlers for PUT and DELETE to llRequestURL and llRquestSecureURL

mdickson:
Per the SL Wiki the Request URL functionality should work for GET/POST/PUT and DELETE.  We were only installing handlers for GET and POST.  This corrects that.


appurist:
We should at some point seek to replace these specific methods with handlers that accept _any_ HTTP method. (There are 8 standard ones defined in [RFC 7231](https://tools.ietf.org/html/rfc7231#section-4),  plus PATCH defined in [RFC-5789](https://tools.ietf.org/html/rfc5789)), and many more registered under the HTTP protocol since then.  Since the HTTP methods are open-ended (not specific), providing a general-purpose HTTP server function should really mean supporting any method (wildcard).  I'd like to see us replace the method-specific handlers (all invoking the same handler anyway) with something that handled wildcard method names.  We should probably at least support accepting HEAD, which is one of only two that are _required_ for servers (HEAD and GET).  All others are optional.

There are dozens of known [registered methods](http://www.iana.org/assignments/http-methods/http-methods.xhtml).  Supporting four of these (probably the most common 4) is a good incremental improvement, and the change looks good.  I'm going to merge it but I'd like us to come back to this in the future with a wildcard approach.

appurist:
Oh by the way, I think the AppVeyor build that failed actually succeeded but was only unable to upload the resulting artifact to @mdickson branch (to be expected).

Issue #439, opened 2018-04-19
Halcyon.sample.ini causes NullReferenceException crash related to AvatarCurrency


To-mos:
The `Halcyon.sample.ini` file is missing a definition that causes a `NullReferenceException` in `OpenSim.Region.CoreModules.Avatar.Currency.AvatarCurrency.Initialize()` adding `[Economy]` to the `Halcyon.ini` config fixes the issue.

appurist:
Good catch! I'll update the halcyon.sample.ini with an [Economy] section and add some of the definitions that we use here.  e.g. something like:
```
[Economy]
    ; CurrencyServer = "http://localhost/currency.php"
    LandServer = "http://localhost/landtool.php"

    ObjectCapacity = 45000

    ; Money Unit fee to upload textures, animations etc
    PriceUpload = 0

    ; Money Unit fee to create groups
    PriceGroupCreate = 0
```

appurist:
I've added pull request #444 to update the ini file, which doesn't fix the NRE but the sample INI should include this section as well.

appurist:
I've added `EconomyBaseAccount` to the sample INI file in #444.

Ana-Green:
Hi, dear people, my dad did try to figure it out but could find it it is about the last picture or snapshot (a bit off topic) it does not show it on 36 on 37 it did now its just black if you log in again any ideas? 

Further as for the [Economy] section is it possible to add Podex?  https://www.podex.com/

Best  Ana Green 

And thanks to Bob or Vin for everything :)

kf6kjg:
Ana, could you post each problem/idea to their own new issues so devs can see and respond to each properly?  Thanks.

Issue #70, opened 2015-11-03
Halcyon.Sample.Ini Economy Section is Missing


Vinhold:
The released sample Halcyon.ini files does not have the Economy section in it. This is the only place where the price values can be set for controlling all uploads, group creation and other settings related to fees and economy controls. While most grids wont need or use money operations this is still the only option open to control access to upload and group creation in a grid if those features are to be restricted to only approved persons. 

It looks like this was the only place to set the default prim limit for any region, though this might only be a redundant entry since the value is set in each region.

[Economy]

```
; 45000 is the highest value that the sim could possibly report because of protocol constraints
ObjectCapacity = 30000
; Money Unit fee to upload textures, animations etc
PriceUpload = 0
; Money Unit fee to create groups
PriceGroupCreate = 0
; This is the account Money goes to for fees.  Remember, economy requires that money circulates somewhere... even if it's an upload fee
;EconomyBaseAccount = <UUID of Account that holds the starting cash>
```

kf6kjg:
Any comment improvements you can suggest? Jims got a PR is for this now: #444

appurist:
As I mentioned in #444, I think many of the entries above (like the UserLevel, Stipend-related, etc) are old options copied from OpenSim but not used in Halcyon. I think some of the comments are very specific to OpenSim concepts that don't apply here, and I don't think we should be attempting to document any other options in the sample INI file unless they are confirmed to still apply and work correctly in Halcyon. I think the ObjectCapacity option should not be used from INI file, and may be ignored in some parts of the code in favor of the expected region.xml file option.

I have added an `[Economy]` section in PR #444, and with the key entries including an unused `EconomyBaseAccount` and comment, along with `PriceUpload` and `PriceGroupCreate`.  That's a good start. 

I think additional entries will require specific source-level investigation to see if they are still appropriate and/or even supported at all.  Given the many higher priorities, I don't think this will happen any time soon, but we could leave this open on the backlog since we really should identify the options supported and document those.

Vinhold:
My point was to have the [Economy] section restored and what items in it that are being used. Anything else is nice if it is known what it does and be better documented if it is supported.

appurist:
Okay that's great, and should be resolved by #444.  And as we discover or make use of any other options that have code behind them, we can add those to the Halcyon.sample.ini as well.

kf6kjg:
#444 has been merged for some time. Closing.

PR #444, opened 2018-04-23
Added reasonable values for the missing [Economy] section for #439 startup exception

appurist:
Doesn't actually fix the null reference exception, that would require another code change, but the default Halcyon.sample.ini file should not ship with the missing section anyway.

### DIFF
--- a/bin/Halcyon.sample.ini
+++ b/bin/Halcyon.sample.ini
@@ -481,16 +481,17 @@
     ;slurl_base = "http://places.inworldz.com/"
 
 [Economy]
-    ; CurrencyServer = "http://localhost/currency.php"
-    LandServer = "http://localhost/landtool.php"
-
-    ObjectCapacity = 45000
-
     ; Money Unit fee to upload textures, animations etc
     PriceUpload = 0
 
     ; Money Unit fee to create groups
     PriceGroupCreate = 0
+
+    ; CurrencyServer = "http://localhost/currency.php"
+    LandServer = "http://localhost/landtool.php"
+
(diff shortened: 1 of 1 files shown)


kf6kjg:
I recommend adding a little more commentary, a la issue #70. Other than that, this has been long on my todo list - good to see it happen!

appurist:
I've added some entries for this section in PR #444 but I see now that my commit was also missing `EconomyBaseAccount` which should be added.

I could be wrong, but I think most of the others in Vin's comment above (like the UserLevel, Stipend, etc) are old options copied from OpenSim but not used in Halcyon. I think some of the comments are very specific to OpenSim concepts that don't apply here, and I don't think we should be attempting to document any other options in the sample INI file unless they are confirmed to still apply and work correctly in Halcyon.  I think the `ObjectCapacity` option should not be used from INI file, and may be ignored in some parts of the code in favor of the expected `region.xml` file option.

My immediate goal for _this_ issue is only to fix the null reference exception in this report, not to complete the documentation for all options. However if there are any glaring exceptions (like the `EconomyBaseAccount` above), we should add those as well, as soon as possible. So for now in this PR, I've removed `ObjectCapacity` from the sample file and added `EconomyBaseAccount`.

Vinhold:
Works for me. As long as the options for the prices and bank account UUID are in, I am good to go. That is what the website uses with the features in the simulator. Max prims is actually set with the region settings. So I have no idea why it was in the [Economy] section.

appurist:
I've added the `EconomyBaseAccount` entry, but I don't think there's any others that are safe to add without further investigation. That may be necessary for #70 but I don't think it's necessary for this one. Ready to go?

Vinhold:
Works for me. One advance at at time as it can be verified. :)

PR #445, opened 2018-04-24
Tell the ServicePointManager to prefer TLS1.2

mdickson:
Set an enum that makes .NET 4.5 prefer TLS 1.2

### DIFF
--- a/OpenSim/Framework/Servers/HttpServer/BaseHttpServer.cs
+++ b/OpenSim/Framework/Servers/HttpServer/BaseHttpServer.cs
@@ -1496,7 +1496,7 @@ namespace OpenSim.Framework.Servers.HttpServer
 
             try
             {
-                ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
+                ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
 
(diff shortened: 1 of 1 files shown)


appurist:
Isn't it true that this is _not_ telling .NET to prefer TLS 1.2, but rather to support _only_ TLS 1.2? 

Wouldn't it be much safer to use: `ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;` so that TLS 1.2 is _added_ to the supported list. The way I understand it is that .NET will automatically prefer the highest/best protocol if available, so that would work just as we want it.  I'm a bit concerned that some websites that used to be supported from Halcyon servers would no longer be supported.  (I know it's crytographically insecure but it could be a change that breaks existing content, and we continue to support HTTP anyway.)

It also seems that the default for `SecurityProtocol` in 4.6.x isn't just `Tls12`, it's `Tls | Tls11 | Tls12`, so _adding_ it to the existing values should make it work as if we upgraded to 4.6.x.

Also, as we move forward in the future, setting it explicitly to `Tls12` would _limit_ the code from using better method, which might become important if TLS 1.2 also eventually becomes insecure.

appurist:
Excellent, merging...

PR #446, opened 2018-05-06
Drop Prebuild

kf6kjg:
It is time.  It took me a lot of work to make the result actually compile* under Ubuntu 16.04 with the FlatPak install of MonoDevelop 7.3.3.

Note I specifically did not attempt to add NuGet dependency management: that's a totally separate project with its own complexities.

I've also not added a bunch of rules to the solution/project files that I think could be added, I just took the output of prebuild and made the edits I needed to make it work.

* Although I had to also remove the `__MonoCS__` compiler checks too, not committed here as they cause other problems.  As such they'll be a separate PR.


kf6kjg:
Looks like I need to fix something though: the AppVeyor build failed because it was unable to find LMV for some reason. Though that should be right there in the bin.

kf6kjg:
AppVeyor is still failing, but on something that should exist under Windows/.NET builds.  Odd.

appurist:
I'll pull the branch anew and see if I have a similar build problem.

appurist:
Looks great. Built without trouble, everything starts and runs and login successful.

Time to clean up the `runprebuild` stuff after this. Woot! \o/

PR #450, opened 2018-05-15
Enable Ubuntu builds on AppVeyor

kf6kjg:
Also brings things up to my latest appveyor work.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,8 +1,19 @@
 version: 0.9.00.{build}
 
-image: Visual Studio 2017
+image:
+- Visual Studio 2017
+- Ubuntu
 
 configuration: Release
+
+platform:
+- x64
+
+pull_requests:
+  do_not_increment_build_number: true
+
+nuget:
+  disable_publish_on_pr: true
 
 skip_commits:
   files:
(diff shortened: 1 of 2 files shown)


mdickson:
The other possible fix is here.....

https://www.nuget.org/packages/System.DirectoryServices.AccountManagement/

Which would remove the need for the hackery since there is now a portable version of the functions being used.

kf6kjg:
It might, would have to be tested.

appurist:
Looks like changing the conditional to `_MONO_CLI_FLAG_` and making use of that resolves the problems. Looks good.

mdickson:
It lets the code build on Linux but the code path is completely non functional. It just always returns false.  Better to take the conditional code out and test with the NuGet library on Linux.   Halcyon isnt functional on Linux as it is now and the current code will continue to compile as use the .NET 4.5 support in Windows.

appurist:
Sorry, there were no comments on this when I merged it, none from either Mike or Ricky (I thought github auto-refreshed the page?) and I confirmed Ricky's intention and readiness to merge privately.

Anyhow, doing it in two pieces is probably for the best anyway. There's an immediate resolution to the impact of trying to support AppVeyor Linux builds, without any impact on the Windows builds.  And it means we also have the option of proceeding now with a separate Part 2 PR with a migration to a NuGet library. I'm somewhat nervous about replacing the System library used currently in Windows with one from NuGet one, but it should resolve the need for platform-specific code.  Once available, this would then require new testing on both Windows and Linux since both are affected.  Mike if you (or Ricky) want to proceed with a second PR, I'd welcome that.

Issue #418, opened 2018-01-24
Money Enforcement on Uploads Inconsistent


Vinhold:
I just installed a new grid to test some other features and made this discovery:
No Economy was put in place with the website (which creates the bank user and UUID setting a value for the world bank in the DB); however the Halcyon.ini did have an incorrect world UUID for the bank account, the account did not exist in the DB. Prices were set for for PriceUpload and PriceGroupCreate in Halcyon.ini.
I was able to create a group with no money in my user account. But I was not allowed to upload a texture.
Money test should be applied for uploads and group creation consistently. I propose this test:

Apply money payment requirement only if there is a Bank account in the DB (valid UUID check set in Halcyon.ini) and world has funds>0 and there is a Halcyon.ini money value >0 assigned to the actions in the [Economy]. Then see if the User has the funds to pay for the transaction.

appurist:
We have been moving to money transfers that indicate success/fail.  See [Mantis 1747](https://bugs.inworldz.com/mantis/view.php?id=1747) (now implemented) and [Mantis 3258](https://bugs.inworldz.com/mantis/view.php?id=3258) which is planned soon. The server code may be recognizing that the bank avatar ID is unable to receive the transfer and failing it in the case of the upload.  (This would actually be a bug in the groups case, however since you cannot create a group from a script it might ignore the failure and proceed.)

I would suggest that you retest this with a **valid** `EconomyBaseAccount` (e.g. create a user named "World Bank" or something and use that UUID. That's the way a grid economy is intended to be set up.  Giving it a valid UUID may allow the upload transfer to succeed.

appurist:
If we fix the inconsistency, to confirm the actual funds transfer in groups as we do for uploads, _neither_ would work with your invalid currency account.  That would fix the inconsistency (and also any hacker windows of opportunity on simultaneous groups creation) and be coded as it should be, but I think this would mean both operations would fail with your current setup. So tests with a valid UUID are definitely in order here.

Vinhold:
My issue statement is that the implementation of upload fee and group creation fee when PriceUpload and PriceGroupCreate settings when the {Economy] are set, is inconsistent. Now I can say it is incomplete and the results are not what is expected. Here are a set of tests run on Halcyon release 0.9.36:

This set of tests is to examine operation of the economy process under selective conditions and record what happens to see if there are any edge cases where it fails to work as expected.

**1. No Economy: This means there are no grid fees set, prices set in Halcyon.ini [Economy]:**
[...]

PR #461, opened 2018-07-13
Fix issue #418: Invoke the code to check and apply a fee on group creation if configured.

appurist:
The code to check and apply a fee on group create has never been invoked. This commit updates the group creation (when using FlexiGroups) to check and apply if configured with `PriceGroupCreate` in the `Halcyon.ini` file.

### DIFF
--- a/OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
+++ b/OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
@@ -1049,11 +1049,14 @@ namespace OpenSim.Region.OptionalModules.Avatar.FlexiGroups
             }
 
             UUID groupID = m_groupData.CreateGroup(grID, name, charter, showInList, insigniaID, membershipFee, openEnrollment, allowPublish, maturePublish, remoteClient.AgentId);
-            if (groupID != UUID.Zero)
+            if (groupID == UUID.Zero)
             {
-                mm.ApplyGroupCreationCharge(remoteClient.AgentId);
-                remoteClient.SendCreateGroupReply(groupID, true, "Group created successfullly");
+                remoteClient.SendCreateGroupReply(groupID, true, "Server error attempting to create group (try again later).");
+                return UUID.Zero;
             }
+
+            mm.ApplyGroupCreationCharge(remoteClient.AgentId);
+            remoteClient.SendCreateGroupReply(groupID, true, "Group created successfully.");
 
             // Update the founder with new group information.
             SendAgentGroupDataUpdate(remoteClient, remoteClient.AgentId);
--- a/OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
+++ b/OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
@@ -1034,7 +1034,21 @@ namespace OpenSim.Region.OptionalModules.Avatar.FlexiGroups
                 remoteClient.SendCreateGroupReply(UUID.Zero, false, "A group with the same name already exists.");
                 return UUID.Zero;
             }
-            
+
+            IScene scene = (remoteClient == null) ? m_sceneList[0] : remoteClient.Scene;
+            IMoneyModule mm = scene.RequestModuleInterface<IMoneyModule>();
+            if (mm == null)
+            {
+                remoteClient.SendAgentAlertMessage("Server configuration error - cannot create group without money module.", false);
+                return UUID.Zero;
+            }
+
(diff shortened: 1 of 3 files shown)


OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
```diff
@@ -1034,14 +1034,29 @@ public UUID CreateGroup(IClientAPI remoteClient, string name, string charter, bo
...
             }
-            
-            UUID groupID = m_groupData.CreateGroup(grID, name, charter, showInList, insigniaID, membershipFee, openEnrollment, allowPublish, maturePublish, remoteClient.AgentId);
 
-            remoteClient.SendCreateGroupReply(groupID, true, "Group created successfullly");
+            IScene scene = (remoteClient == null) ? m_sceneList[0] : remoteClient.Scene;
+            IMoneyModule mm = scene.RequestModuleInterface<IMoneyModule>();
+            if (mm == null)
+            {
+                remoteClient.SendAgentAlertMessage("Server configuration error - cannot create group without money module.", false);
```
kf6kjg: This introduces a new semantic: previously the absence of a money module still allowed group creation.  I suggest restoring that semantic.

OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
```diff
@@ -1034,14 +1034,29 @@ public UUID CreateGroup(IClientAPI remoteClient, string name, string charter, bo
...
+                remoteClient.SendAgentAlertMessage("Unable to create group. Insufficient funds.", false);
+                return UUID.Zero;
+            }
+
+            UUID groupID = m_groupData.CreateGroup(grID, name, charter, showInList, insigniaID, membershipFee, openEnrollment, allowPublish, maturePublish, remoteClient.AgentId);
+            if (groupID != UUID.Zero)
+            {
+                mm.ApplyGroupCreationCharge(remoteClient.AgentId);
+                remoteClient.SendCreateGroupReply(groupID, true, "Group created successfullly");
+            }
```
kf6kjg: No error to the user if the group failed to be created?

OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
appurist: While it is true the absence of the money module allowed it previously (on this one operation), that is because this one operation was completely missing all money-related code/checks.  (Sending an IM to a user also succeeds with no money module because of course that operation is currently free...)

As far as I understand things, the money module is not optional, there are many places that would crash or fail if one was not present.  (One example discovered below is that mesh uploads will crash.)

More importantly, this suggestion would make it inconsistent with the application of the money balance checks used on other operations (e.g. in `HandleUDPUploadRequest` and `NewAgentInventoryRequest` which both have code to return an error to the user if there is no money module.  Clearly the intention is to fail an operation that has a fee if there is no money module available.

However, I notice that `NewAgentInventoryRequest` (the mesh upload case) will crash with a null reference exception just before it attempts to report the insufficient funds if there's no money module defined.  That said, crashing before performing the operation still means it won't succeed when there is no money module.

I modeled this change after the other cases, thus it has the same failure result if there's a fee and no money module installed.  I don't think we should be changing that behavior for this specific new check, just because there wasn't any check at all previously.

OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
appurist: Good point. I improved the previous behavior, which was to always report the group creation as a success, even when it failed, by making that result conditional on actual success. I didn't add a new message for the failure case, and maybe that would be a good improvement here.

appurist:
I'm going to add an error response to the user in the event of a group creation failures. I'm also going to take a look at skipping all of this code (first comment) if the operation is free.

OpenSim/Region/OptionalModules/Avatar/FlexiGroups/FlexiGroupsModule.cs
kf6kjg: That makes sense then: always helps to see the larger picture.

appurist:
Unfortunately it's not possible to precheck if a fee applies to group creation because it is the money module that owns that data (that's why we're trying to get a reference to the money module here in the first place, to do that check). However, I've added the user message for the failure case (and fixed the spelling error in the success case!).

appurist:
I'm going to fix that mesh upload null references as part of this PR since it's basically the same code. One more commit coming.

PR #465, opened 2018-08-05
Matched assembly versioning to actual versioning

kf6kjg:
Helps keep things understandable.


appurist:
This will break the auto-build numbering.

kf6kjg:
Hence it has the change for appveyor's automatic building built in.  Unless
you mean some other system?

appurist:
Last time I checked, the automatic build numbers in Visual Studio on support the * in the third place. The * in the fourth place was something like seconds since midnight, which would mean non-incrementing build numbers. This is why the AssemblyVersion did not include the leading "0." and restoring in will cause the build number to be somewhat random.

PR #466, opened 2018-08-07
Moved versioning to a T4 template

kf6kjg:
This allows the correct build sequence number, unlike my last attempt.

PR #467, opened 2018-08-10
Fixed #456: Region settings are now saved on load from OAR

### LINKED ISSUE #456: Loading Oar file, terrain settings are not saved to regionsettings table.
Example case:
In a beta or test grid create a new region. Apply a new land terrain to the region. Put in custom terrain textures and blending settings for it. Make it really stand out different from the default region textures.
Save the region as an oar file.
Load the saved oar file into a production Halcyon grid.
The region will load and the custom textures will apply. You can even see them in the viewer region terrain window all correct. Check the region's regionsettings table record, it will not have any of the texture UUIDs or settings to match. The data is not saved to that record.
Shutdown and restart the region. The terrain textures and blending levels will all be reverted to the defaults seen in the regionsettings table record and the custom ones loaded will all be gone.
It does not matter if the region has any content in it or not. This affects only the terrain textures and blending levels.

kf6kjg:


kf6kjg:
Build failure is on the Ubuntu side and is fixed by #466.  Waiting on testing by Vinhold.

appurist:
If I recall correctly, I think I only "recently" simplified that settings interface with the Save(), so that's why it may have been missing (it didn't exist when the region oar load code was written). 

Looks good to me, ready for merge.

